PR #9, opened 2019-03-08
Support updates for Issues/1, add additional scan options

cdcabrera:
## What's included

* fix(toolTip): converted simpleToolTip to Tooltip 
  * converted simpleToolTip to Tooltip
  * encases both popover and tooltips
  * generates a random id if not provided
* fix(lodash, redux): update lodash, redux imports 
  * update lodash, redux imports across remaining components
  * updated listStatusItem and sourceCredentialsList component logic
  * relocated createScanDialog in prep for adding cli options

### Notes

* This is a behind the scenes set of updates. Within the UI there should be no discernible difference to an end user.

1. update the NPM packages with `$ npm install` or `$ yarn`
1. `$ npm test` or `$ yarn test`
1. update the NPM packages with `$ npm install` or `$ yarn`
1. `$ npm run test:dev` or `$ yarn test:dev`

## Updates issue/story
#1


cdcabrera:
We agree on the lodash-es, replacing the imports should thankfully be pretty straight forward.

However, we currently don't gain anything on the code splitting side of things because of the older version of create react app we're on, 1x. Since we plan on keeping create react app around the switch over to lodash-es would correlate nicely as part of a final packaging solution in a series of fixes and updates aimed at fixing how we handle Redux, and life cycle within the components themselves. 

Fixing those bits first, then upgrading the framework, lodash and other build packages we should see a noticeable improvement in overall GUI response times

PR #27, opened 2019-03-22
feat(styling): issues/6 refresh for brand logos, styling

cdcabrera:
## What's included

* feat(styling): issues/6 refresh for brand logos, styling
* DEV scss generic rename

1. update the NPM packages with `$ npm install` or `$ yarn`
1. `$ npm test` or `$ yarn test`
1. update the NPM packages with `$ npm install` or `$ yarn`
1. `$ npm run test:dev` or `$ yarn test:dev`

## Example
Screen Shot 2019-03-22 at 6 44 48 PM

Screen Shot 2019-03-22 at 6 45 02 PM

Screen Shot 2019-03-22 at 6 45 15 PM

Screen Shot 2019-03-22 at 6 45 30 PM

Screen Shot 2019-03-22 at 6 45 37 PM

## Updates issue/story
Updates #6


jlprevatt:
Premise from earlier chats with Dan:
- The name of the component is "product discovery tool" with no caps because it's a component.
- However, Dan has also stated that we can call this thing "discovery" and his preference is that the shorter name would catch on.
- From my past experience, it's kinda really weird that a *component* would get both a long name and short name, but I'm trying to achieve zen with that and roll on.

However, the changes here are introducing a third name, "product discovery." I think three names are too much. So, we need to decide how and where "product discovery tool" will live on screen, and how and where and if "discovery" will live on screen--even in places that are not space-constrained, which is the usual requirement for using the short name on screen.

Here are some options:

1. Use "product discovery tool" (with any "the" or other needed text to smooth it out where needed) everywhere. Doesn't meet Dan's goal to get the shorter "discovery" to catch on.
1. Use "discovery" in most places, such as the places it appears in large font (above the login fields, on the welcome page, in the window title) and introduce the full name in the running text (paragraphs) such as "Welcome to discovery, the product discovery tool. ..."
1. A hybrid that *might get too confusing.* Such as "discovery" in most places and inserting "product discovery tool" like the previous suggestion, but spell out "product discovery tool" in the window title.
1. Others that I haven't thought of yet....

dozdowski:
Like Jamie said, my main take it, "We follow Brand's rules about what we can and cannot say/do/use - but we should guide the creation of a tool that you can use and talk about without it getting weird and/or stupid."

In my opinion, we should ignore the "tool" bit and leave "product discovery" in three places:

1. The main login screen
2. The first line of the login screen's intro copy
3. The top left banner

In all other places, we should drop back to "discovery" which is an informal and familiar enough name to facilitate conversation around usage of the tool.  Nobody says "Ansible runner" when referring to that component, they just say "runner." My goal is to push "discovery" as our equivalent token.

cdcabrera:
### Review updates:
#### Sources
Screen Shot 2019-03-26 at 1 47 36 PM


#### Scans
Screen Shot 2019-03-26 at 1 47 48 PM


#### Credentials
Screen Shot 2019-03-26 at 1 47 21 PM

public/index.html
```diff
@@ -6,7 +6,7 @@
     <meta name="theme-color" content="#000000">
     <link rel="manifest" href="%PUBLIC_URL%/manifest.json">
     <link rel="shortcut icon" href="%PUBLIC_URL%/favicon.ico">
-    <title>Entitlements Reporting</title>
+    <title>product discovery</title>
```
jeff-phillips-18: Not sure I've ever seen a product title that wasn't init-capped. If this is the title shouldn't it be (throughout)?

src/components/credentials/credentialsEmptyState.js
```diff
@@ -8,11 +8,11 @@ const CredentialsEmptyState = ({ onAddCredential, onAddSource }) => (
     <Row>
       <EmptyState className="full-page-blank-slate">
         <EmptyState.Icon />
-        <EmptyState.Title>Welcome to {helpers.RH_BRAND && 'Red Hat'} Entitlements Reporting</EmptyState.Title>
+        <EmptyState.Title>Welcome to {helpers.RH_BRAND && 'Red Hat'} discovery</EmptyState.Title>
```
jeff-phillips-18: Causes an extra space `Welcome to__discovery` when there is no RH_BRAND.

public/index.html
cdcabrera: @bclarhk

src/components/credentials/credentialsEmptyState.js
cdcabrera: indeed, it just happens to take advantage of the older html and browser behavior where double spaces display as a single, plus the screen reader seems to skip it.

i suppose it could also be formatted as `{helpers.RH_BRAND && 'Red Hat '}discovery `

cdcabrera:
... looks like we still need the ico file in there too

src/components/credentials/credentialsEmptyState.js
cdcabrera: went ahead and updated to remove the spacing

public/index.html
bclarhk: it's the new branding standard for non-official products

cdcabrera:
Updated screenshots for login/logout
Screen Shot 2019-03-27 at 4 24 12 PM


Screen Shot 2019-03-27 at 4 24 23 PM

Issue #38, opened 2019-04-03
Content services copy integration


cdcabrera:
## Feature description
As a developer it would be helpful to be integrated with a documentation or copy based repository so that copy string changes can be made independent of the UI code.

### Describe the solution you'd like
- Option 1: NPM build package reference to a copy repository. Updated strings get checked in to the project manually.
- Option 2: Full CI integration of the NPM build package. Jenkins would run an additional step where the appropriate copy build script would generate the copy strings. This process would possibly require a baseline or fallback set of strings.
## Acceptance Criteria
- [x] Verify the strings currently (before this update) within the copy repo are being used in ui repo.
- [x] Verify README or CONTRIBUTION in the UI and copy repos have been updated to explain this process

___

## Additional context

- Copy strings would be held with JSON resource. The resource could be either loaded through an Ajax request (local files with the ability to expand to an external service in the future), or created as an import module within the app. 
  - This repo uses the Ajax way of handling strings, https://gitlab.com/cloudigrade/frontigrade/issues/90
- Changes to UI would involve adding keys/IDs and implementing a i18n package

cdcabrera:
Baseline string placement ideas
- Deep scan tooltip, scan dialog/modal
- Credential, `View Credential` hover tooltip

Directory placement
- use `dist/ea_out`

cdcabrera:
As of 20190703...

Quipudocs NPM structure has stabilized, and is ready for NPM consumption.

jlprevatt:
As part of this issue, we should consider how we are going to add access to the documentation from the ? menu in the GUI. Such as...

- Two different options (and decide on labels) for install and user documents

- A single "Documentation" or "Help" or whatever menu option with the install and user labels as suboptions

Issue #18, opened 2019-03-12
User Story: Update framework and syntax


kholdaway:
@cdcabrera commented on [Mon Jan 28 2019](https://github.com/quipucords/quipucords/issues/1791)

### Is your feature request related to a problem?
React framework and general optimization updates (Or making the GUI devs lives easier by making this repository reflect updated knowledge for how we structure code bases). 

Currently the GUI is fixed on a specific version of React. In order to upgrade various syntax and React life cycle methods need to be altered. 

### Describe the solution you'd like
Behind the scenes GUI code updates:
- Update services to remove class based structure
- Update redux structure in an effort to simplify state
  - evaluate selectors 
  - update toolbar filters
- Update legacy React life cycle methods within components
  - this should bring with it a reduction in component complexity
- Expand unit tests and coverage to promote GUI code stability, and relax the need to always require another UI dev resource in addition to the QE team reviews
  - expose coverage and build aspects on the GUI Readme, aim for a minimum of 70% + coverage
___

## Acceptance Criteria:
- [x] Services should no longer contain class based structure
  - separate helper functions into `common/helpers`
- Clean up Redux structure, should include 
  - [x] removing unused types, actions, reducer checks
  - ~clean up lodash usage within redux, evaluate `lodash-es`~ until we can move to a newer version of the react scripts going to delay this
  - [x] unit tests expanded
- [x] Verify that users see no discernible negative difference within the GUI. increases in response and render time may be a positive outcome
- [x] Verify that GUI unit tests are expanded and achieve a minimum of 70% coverage
  - [x] Component testing 
  - [x] State / Redux testing
  - [x] Services testing
- [x] Verify that legacy React life cycle methods have been removed... searching in the code base for `componentWillReceiveProps` should return 0 results 

## Related issues
#1520

cdcabrera:
As of 20190529... 

There is one last grouping of components that retain the legacy lifecycle methods... Credentials components. We may end up doing a quick conversion then iterate on it instead of our normal in-depth conversion. This would let us close this ongoing issue down.

cdcabrera:
As of 20190703...

There is still one last grouping of components that retain the legacy lifecycle methods... Credentials components.

cdcabrera:
Marking this one finally resolved as of 20220628 as part of the prep for the segue to patternfly 4
- updated to react 17
   - one or two legacy lifecycles are still in there, but they should be removed as part of the pf4 refactors come through
- npm packages updated
[...]

Issue #50, opened 2019-05-09
Add a Makefile to clean up Docker


cdcabrera:
## Feature description
As a dev and/or tester I would like the ability to clean up my review environment consistently.

### Is your feature request related to a problem?

Yes, there appears to be some caching going on during PR reviews, when updates happen they're only partially being reimplemented when the Docker review image is "rebuilt"

### Describe the solution you'd like

Add a `Makefile`, for consistency with Quipucords proper, that allows clean up
___

## Acceptance Criteria
- [ ] Verify that the Makefile successfully resets/wipes out the previous "review" image

PR #51, opened 2019-05-17
feat(i18n): issues/38 i18n integration, link to quipudocs

cdcabrera:
## What's included

- feat(i18n): issues/38 i18n integration, link to quipudocs
  * build, npm scripts for generating pot file and building from quipudocs
  * build, dotenv paths and expire added, default locale
  * build, make POT available to quipudocs team
  * app, i18n, component integration
  * i18n component to display strings
  * userService, setup base for pulling locale
  * redux, userActions, userReducer, locale into user session
  * aboutModal, set up baseline for string replacement
- DEV lint packages updated, included html hint
  * build, added htmlhint, prep quipudocs help files
  * build, integration test snapshot for locales

### Run the Review UI
1. clone the repo, create a branch, pull the changes
1. cd into the repo, update the NPM packages with `$ yarn`, if you don't have yarn installed you can use brew or `$ npm install yarn -g`, next
1. make sure Docker is running, then
1. `$ yarn start:review` ... this should compile the image, then run the container with the latest changes
1. From there the labels inside the About modal are currently setup to either pass a translation string from '[repo]/public/locals/en.json' or use the default strings and pass them through. This file is a placeholder, and is expected to be an empty JSON object.
  - In the browser console an AJAX call to `en.json` should register

## Example

Screen Shot 2019-05-17 at 5 01 23 PM

Screen Shot 2019-05-17 at 5 08 20 PM

## Updates issue/story
#38 relates to https://github.com/quipucords/quipucords/pull/1921
#18


.env.test
```diff
@@ -6,6 +6,12 @@ REACT_APP_AUTH_TOKEN=csrfspoof
 REACT_APP_AUTH_HEADER=X-CSRFToken
 REACT_APP_AJAX_TIMEOUT=60000
 
+REACT_APP_CONFIG_SERVICE_LOCALES_DEFAULT_LNG=en
+REACT_APP_CONFIG_SERVICE_LOCALES_DEFAULT_LNG_DESC=English
+REACT_APP_CONFIG_SERVICE_LOCALES=/locales/locales.json
+REACT_APP_CONFIG_SERVICE_LOCALES_PATH=/locales/{{lng}}.json
+REACT_APP_CONFIG_SERVICE_LOCALES_EXPIRE=604800000
+
```
jeff-phillips-18: Any way to have a `.env.base` or something where all these can get base settings from so that we only need the overrides in these files?

src/components/aboutModal/__tests__/__snapshots__/aboutModal.test.js.snap
```diff
@@ -68,64 +68,18 @@ Object {
...
-        type="fa"
-      />
-    </Button>
-  </div>
-</AboutModal>
+  t={[Function]}
+  tReady={true}
+  uiVersion="0.0.0.0000000"
+  username="lorem"
+/>
```
jeff-phillips-18: why such a big change to the snapshot? I would've thought just some minor string changes.

src/components/i18n/i18n.js
```diff
@@ -0,0 +1,42 @@
...
+      react: {
+        useSuspense: false
+      }
+    });
+
+/**
+ * FixMe: the "I18nextProvider" appears to have timing issues, reevaluate this alt implementation
+ */
+const I18n = ({ children, locale }) => {
+  i18nInit(locale);
```
jeff-phillips-18: Do we know if this is getting called multiple times (if app re-renders)? Should it or should we track if it has already been initialized?

.env.test
cdcabrera: looks like we could try using just using a `.env` that would combine `.test` `.production` and `.staging` I'll test it out may make that a follow up task since it's more related to the build, can combine it with #50

src/components/aboutModal/__tests__/__snapshots__/aboutModal.test.js.snap
cdcabrera: I'm glad you asked =) 

This is all the the translation component wrapper. Because we decided to pass the component through the "translation" piece here
- https://github.com/quipucords/quipucords-ui/pull/51/files#diff-233d6105e91463bd8457d5414ada705dR10

the translation "connected" aspect appears to not fire the same as the original connect. debated on whether to place a "test" condition check in the `index.js` file above. 

```
const connectTranslate = (mapStateToProps, mapDispatchToProps) => component =>
  connect(
    mapStateToProps,
    mapDispatchToProps
  )((process.env.REACT_APP_ENV !== 'test' && withTranslation()(component)) || component);
```

Eventually we opted for the sans-test condition, but willing to put it back in there, it would bring back expected testing behavior, thoughts?

src/components/i18n/i18n.js
cdcabrera: Yeahhhhh, was noticing some speed differences here. Went back and double checked we should probably change it to be more in line with a normal life cycle method.
### before
What happens currently during the firing...
- Initial app view load it fires twice, thats expected, the second time being the fulfilled locales ajax call
- View load, it'll fire once
- Subsequent navigation, page loads, modal edits do not affect it since the views aren't being reloaded
### after the below change
What happens after we switch over to the below
- Initial app view load it fires twice, as expected when the props refresh from the ajax call, that's it, no subsequent calls on view change... also I feel silly for missing that, I'll make the change =)
```
class I18n extends React.Component {
  componentDidMount() {
    const { locale } = this.props;

    i18nInit(locale);
  }

  componentDidUpdate(prevProps) {
    const { locale } = this.props;

    if (locale !== null && locale !== prevProps.locale) {
      i18nInit(locale);
    }
  }

  render() {
    const { children } = this.props;
    return <React.Fragment>{children}</React.Fragment>;
  }
}
```

src/components/aboutModal/__tests__/__snapshots__/aboutModal.test.js.snap
jeff-phillips-18: If you think the snapshot accurately depicts what is being set and it is testing what we want to test, then I am fine with it.

cdcabrera: Sure thing, after going over it again. I think it's a good call out, going to roll the `test` condition version. Not being able to expect `component.dive()` to behave as it did before in a snapshot is just hiding/masking/abstracting things, and something that could lead to confusion again if/when translation is expanded

Issue #91, opened 2022-06-28
Component: Alert to PF4 components/styling


jenny-s51:
## Acceptance criteria
- Alert component to PF4

Components affected:

- [x] createCredentialDialog.js
- [x] createScanDialog.js
- [x] credentials.js
- [x] scans.js
- [x] sources.js
- [x] authentication.js

## Assumptions and questions
- there will be styling differences associated with Alert
- Confirm with UX on styling alterations. You can also do this while working or as WIP/Draft PR.
- attempt to convert the strings towards the `translate` method. example can be found here - https://github.com/cdcabrera/quipucords-ui/commit/b3c8eaa48d844f46ca6f6dd44c9cd0c84f5dcc52

### Related issue
[DISCOVERY-148](https://issues.redhat.com/browse/DISCOVERY-148)

PR #115, opened 2022-07-06
Refactor(toastNotificationsList.js): Use PF4 Alert instead of Toasts

### LINKED ISSUE #109: Component: ToastNotificationsList to PF4
While working on #91 I noticed that redux manages the toast notifications in a dispatch call.

### Acceptance criteria
Use `Alert` components instead of Toasts within ToastNotificationsList.

## Assumptions and questions
- Related to #91 conversion of Alerts to PF4 alerts - PF3 toast notifications have been replaced with PF4 timeout alerts.

### Related issue
[DISCOVERY-148](https://issues.redhat.com/browse/DISCOVERY-148)
#119

jenny-s51:
## What's included

...
Updates to toastNotificationsList -> now using `AlertGroup` and `Alert` instead of `ToastNotificationList` and `TimedToastNotification`.

## How to test

Test "Create", "Scan", "Edit", "Delete", etc. throughout the UI to view updates to the toast alerts. 

## Example

...

Before:
Screen Shot 2022-07-06 at 3 22 54 PM
Screen Shot 2022-07-06 at 3 22 46 PM
Screen Shot 2022-07-06 at 3 22 32 PM

After:

Screen Shot 2022-07-06 at 3 58 57 PM
Screen Shot 2022-07-06 at 3 57 48 PM
Screen Shot 2022-07-06 at 3 54 53 PM

## Updates issue/story

Extends #91 , closes #109


src/styles/app/_overrides.scss
```diff
@@ -11,6 +11,10 @@
   }
 }
 
+.pf-c-alert-group.qpc-alert-group.pf-m-toast {
+  width: fit-content;
+}
+
```
cdcabrera, line 17: required:
lets look at moving this over to its own file like `_toastNotifications.scss`. we've been debating if we should get rid of the `override.scss` and `patternflyoverride.scss` files... since technically it's all overrides =)

src/components/toastNotificationsList/toastNotificationsList.js
```diff
@@ -21,46 +21,41 @@ class ToastNotificationsList extends React.Component {
   };
 
   render() {
-    const { toasts, paused } = this.props;
+    const { toasts } = this.props;
+    const timeout = 8000;
 
     return (
-      <ToastNotificationList>
+      <AlertGroup isToast className="qpc-alert-group">
```
cdcabrera, line 28: required:
update the css className to `quipucords-alert-group`, or attempt something fancy `quipucords-toast-notifications__alert-group` in the event we need to add classes

src/components/toastNotificationsList/toastNotificationsList.js
```diff
@@ -21,46 +21,41 @@ class ToastNotificationsList extends React.Component {
   };
 
   render() {
-    const { toasts, paused } = this.props;
+    const { toasts } = this.props;
+    const timeout = 8000;
```
cdcabrera, line 25: required:
convert hardcoded timeout to a `prop` with a `default` of `8000`... bit easier than having to figure out where that number is buried, and if we do switch to another method of providing that number bit easier to update

optional:
consider making this a dotenv parameter in `.env` maybe something like
```
REACT_APP_TOAST_NOTIFICATIONS_TIMEOUT=8000
```
but then you'd also need to bring it in through `./src/common/helpers.js` annnnnd now it's starting to move into refactor land

cdcabrera (changes requested):
Good first round!

Looks like there's some behavior issues with the new updates. It doesn't appear to be removing the old toasts. 

We keep adding credentials and it keeps showing the prior ones. The old behavior was to remove the toasts from state automagically, so we need to restore that ability.


Screen Shot 2022-07-06 at 5 31 02 PM (2)

bclarhk (approved):
Same css text alignment issues, but otherwise LGTM!

src/styles/app/_overrides.scss
jenny-s51: Good idea! Updated 👍

src/components/toastNotificationsList/toastNotificationsList.js
cdcabrera: dotenv params come across as strings, we just need to convert it towards an int... in helpers go ahead and add this wrapper... the `undefined` is there so you don't receive `NaN`
```
const TOAST_NOTIFICATIONS_TIMEOUT = Number.parseInt(process.env.REACT_APP_TOAST_NOTIFICATIONS_TIMEOUT, 10) || undefined;
```

jenny-s51: Thank you! Worked like a charm

src/components/toastNotificationsList/toastNotificationsList.js
```diff
@@ -21,46 +25,44 @@ class ToastNotificationsList extends React.Component {
   };
 
   render() {
-    const { toasts, paused } = this.props;
+    const { toasts, timeout } = this.props;
 
     return (
-      <ToastNotificationList>
+      <AlertGroup isToast className="quipucords-toast-notifications__alert-group">
```
cdcabrera, line 31: required:
lets update these things
- add `isLiveRegion` to AlertGroup
- flip from `{toasts && toasts.map...` towards `{toasts?.map...`
- title attribute, remove the unnecessary nullish coalescing operator, replace it with an OR `||`... nullish coalescing has its time and place, but old skool truthy/falsy works just as well, especially in the scenario of someone passing an empty string
- lets be consistent with our ternary... replace `toast.header && toast.message ? ... : ''` with the format you used in the title attribute, `(toast.header && toast.message) || ''`

cdcabrera (changes requested):
Looking good so far on the initial steps! Lets handle the requested updates, and we'll move to the next steps with this PR.

I'll look at adding a unit test to this PR to help us work through the process and visualize the changes.

src/components/toastNotificationsList/toastNotificationsList.js
```diff
@@ -21,46 +25,51 @@ class ToastNotificationsList extends React.Component {
...
-                </TimedToastNotification>
-              );
-            }
-
-            return null;
-          })}
-      </ToastNotificationList>
+      <AlertGroup isToast className="quipucords-toast-notifications__alert-group">
+        {toasts?.map(toast => {
+          if (!toast.removed) {
```
cdcabrera, line 33: required:
we need to consume the redux state with the updated toasts. we can leverage the PF timeout attribute's use of `undefined`/`boolean`/`number`

```
...removed ){
return (
  <Alert
    ...
    timeout={!toast.paused && timeout}
```

this update combined with the requested reducer updates should get the toast pause/resume working again. you can take look by setting up a bunch of scans in a row from the scans view

src/components/toastNotificationsList/toastNotificationsList.js
```diff
@@ -21,46 +25,51 @@ class ToastNotificationsList extends React.Component {
...
-                    {toast.message}
-                  </span>
-                </TimedToastNotification>
-              );
-            }
-
-            return null;
-          })}
-      </ToastNotificationList>
+      <AlertGroup isToast className="quipucords-toast-notifications__alert-group">
```
cdcabrera, line 31: required:
can't seem to find a role on the source of AlertGroup, but we do see `aria-atomic` being set on AlertGroup related to the use of `isLiveRegion`. confirm with accessibility we don't need to drop a role attribute on AlertGroup, or the `isLiveRegion` on AlertGroup too

jenny-s51: We have two PF4 `AlertGroup` demos that demo the use of `isLiveRegion` ! Though neither of them use a `role` attribute on `AlertGroup`, they do use an `aria-label` on `Alert`... should we add those @cdcabrera ?

src/redux/reducers/toastNotificationsReducer.js
```diff
@@ -42,18 +42,12 @@ const toastNotificationsReducer = (state = initialState, action) => {
 
     case toastNotificationTypes.TOAST_PAUSE:
       return {
-        ...state,
-        ...{
-          paused: true
-        }
+        ...state
```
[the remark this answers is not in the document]
jenny-s51: Ah thank you for this suggestion! This is a really smart approach to handling this at a toast level rather than at the component level!

src/components/toastNotificationsList/toastNotificationsList.js
cdcabrera: per our accessibility team..."[using the AlertGroup `isLiveRegion` with `isToast` is preferable when the region/component is maintained in a page. the `isLiveRegion` for the Alert component is not as reliable with certain types of screen-readers, and it is not recommended.]" ... this is one of those the docs need to be updated on pf so checking with the UXD accessibility team is the path

jenny-s51: Ah, I see. Now I realize since we have a specific use-case where we are using the toast group (no PF4 react demo for that). Checking with accessibility in those situations is a good thing to keep in mind - will definitely plan to do that moving forward 👍

src/redux/reducers/toastNotificationsReducer.js
```diff
@@ -41,20 +40,20 @@ const toastNotificationsReducer = (state = initialState, action) => {
...
-        ...{
-          paused: false
-        }
-      };
+      const resumedToasts = [...state.toasts];
+      const resumedToastIndex = state.toasts.indexOf(action.toast);
+      if (resumedToastIndex > -1) {
+        resumedToasts[resumedToastIndex].paused = false;
+      }
+      return { ...state, ...{ toasts: resumedToasts } };
```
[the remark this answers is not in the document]
jenny-s51: Thank you! Yes, I was using the ESLint plugin, but seems it's not reliable to emulate our existing styles, so moving forward, I will try to sync up the diffs if there are any unexpected formatting changes!

cdcabrera (approved):
lgtm! i'll merge in a few mins, much appreciated @jenny-s51 !

Issue #96, opened 2022-06-29
Component: Grid to PF4


cdcabrera:
## Acceptance criteria
- Convert grid instances to PF4 component

## Assumptions & questions
- Minimal styling alterations. There may be NO noticeable diffs
- Is used extensively, search for instances and replace

### Related issue
[DISCOVERY-149](https://issues.redhat.com/browse/DISCOVERY-149)

Issue #104, opened 2022-06-29
Component: Form components to PF4


cdcabrera:
## Acceptance criteria
  form components to PF4
  - [x] create a `component/form/` directory and place the following components
     - [x] `formGroup` component - replaces current formField component @jenny-s51 
     - [x] `textInput` component - ported from toolbar work @cdcabrera 
     - [x] `checkbox` component - [ported from Curiosity](https://github.com/RedHatInsights/curiosity-frontend/blob/main/src/components/form/checkbox.js) @jenny-s51 
     - [x] `textArea` component - wrapper for text area @jenny-s51 
     - [x] `radioButton` component - wrapper for radio button @jenny-s51 
     - [x] `formHelpers` - ported from toolbar work - @cdcabrera 

## Assumptions and questions
- ~Most likely to be no visual updates~ larger fields
- It may be easier to flip components while converting other components
- Forms within this app were started being converted over to using a "form state" component for context. It may be easier to do this last

### Related issue
[DISCOVERY-153](https://issues.redhat.com/browse/DISCOVERY-153)

PR #126, opened 2022-07-19
Refactor(Grid): convert Grid instances to PF4 

jenny-s51:
## What's included

...
Towards #96 - I did a bunch of investigation here and a lot of testing/modifying code -- it looks like we won't be able to convert a lot `Grid` occurrences without some pretty heavy refactoring... probably makes sense to update #96 with this info and then figure out the best approach to divide and conquer all those instances! 

In some the create/edit modals we are using `Grid` to apply similar formatting as a horizontal `Form` component, so we need do some refactoring from `Grid.Row`, `Grid.Column` to use the `isHorizontal` variant of `Form` instead of this `Grid`-based approach. I tried to do this to see if it would work out of the box, and it doesn't, so I made a list of things we can start with below. 

Some things we'd need to refactor:

- Remove instances of `Grid` within modals and just use `Form`
- `Form.FormGroup` to `FormGroup`
- `Form.FormControl` to `FormControl`
- Setting `label` on all `FormGroup`s instead of using  `renderFormLabel()`

- Currently investigating `Grid.Row` syntax within some select menus... 

## Example

...
Before:
Screen Shot 2022-07-19 at 5 11 05 PM
Screen Shot 2022-07-19 at 5 11 14 PM

After:
Screen Shot 2022-07-21 at 4 19 24 PM

Screen Shot 2022-07-21 at 4 18 31 PM


cdcabrera:
@jenny-s51 

The grid work has the appearance of needing to be performed as a clean up task. 

That task would be associated with larger grouping components, which in some cases would cause `Grid` to simply be removed without any issue. For instance with one or two of the Empty-State conversions we removed grid entirely.

The preferred method, moving forward, for grid conversions is to
- **Either** work at a slightly higher level, like in the case of this PR, the table/listing conversions first, then work on grid as a follow up
   - Mostly so we don't end up removing grid entirely after we spent effort converting it, 
   - Or using a completely different component like `<List />`, which seems more appropriate for this PR
   - And partially because reusing the old css classes to define how the padding/margins/spacing inside a PF4 Grid behaves inside PF3 work would also, most likely, go away with any PF4 parent component conversion
- **Or** work below grid, like buttons, smaller components



As for the form work, that seems to highlight something similar to the Empty-State work, and grid could simply be considered a removal clean up task in that case. This work has a story #104  and Discovery-153

- Form.FormGroup to FormGroup
- Form.FormControl to FormControl
- Setting label on all FormGroups instead of using renderFormLabel()

As for select menus, I'm in the process now of converting both select lists, dropdowns, and the toolbar components now. If there's something related to those conversions please send it over. However, we haven't noticed anything to do with `grid.row` so we're curious where this is located

jenny-s51:
Thank you for your helpful feedback and clarification @cdcabrera ! Implementing this grid work as a clean up task definitely makes sense and sounds like a good plan.

I'm happy to update this PR with whatever changes you think are best! Just wanted to get a green-light from you first before swapping components around. I think using a `List` instead would probably make sense in a few places, specifically in the screenshots above!

`dropdownSelect.js` is currently implemented using this `Grid.Row` and `Grid.Col` syntax -- https://github.com/quipucords/quipucords-ui/blob/master/src/components/dropdownSelect/dropdownSelect.js#L168-L175

Looks like it can ultimately be refactored to [the multi variant of select!](https://www.patternfly.org/v4/components/select#multiple).

src/components/scanHostList/scanHostList.js
```diff
@@ -88,7 +88,7 @@ class ScanHostList extends React.Component {
 
     return (
       <div className="quipucords-infinite-results">
-        <Grid fluid onScroll={this.onScrollList} className="quipucords-infinite-list">
+        <Grid onScroll={this.onScrollList} className="container-fluid quipucords-infinite-list">
```
cdcabrera, line 91: required:
- lets remove this change from this PR ... reason it's actually a wrapper around Grid... so grid inside grid inside grid... very meta
- OR we can attempt to include the `Grid` changes for the files that call this component, `ScanListItem.js` and `ScanHostList` ... but since we plan on potentially using Grid updates as a clean up task after parent components... it may be easier to just forgo this change for now

But If we keep this file the changes required are...
- same comment about converting the Grid to `<List isPlain` and `<ListItem.`
- throw in the optional chain on `hostsList?.map` for consistency

question:
- possibly apply a similar but NEW `max-height` css class to all lists in this PR, dependent on question to @bclarhk

cdcabrera (changes requested):
- Update grid to list in a couple of spots
- Possibly forgo the grid update on scanhostlist since it's a wrapper of wrappers, plus has a max-height setting we maybe should apply to other lists
- Update unit test snapshots

src/components/credentials/credentialListItem.js
```diff
@@ -127,19 +128,17 @@ class CredentialListItem extends React.Component {
       case 'sources':
         (item.sources || []).sort((item1, item2) => item1.name.localeCompare(item2.name));
         return (
-          <Grid fluid>
+          <Grid className="container-fluid">
```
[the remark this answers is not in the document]
jenny-s51: Yeah I think having the tooltip on the text makes more sense! The larger area around it would help a user toggle it more easily. @bclarhk what do you think of wrapping the tooltip around the string content instead of the icon?

src/components/scanHostList/scanHostList.js
jenny-s51: Thank you for your helpful feedback here - pushed some updates to the files you mentioned. Let me know if this is what you had in mind... might not be _exactly_ 1:1 since the `icon` prop on PF4 `List` has different styling now!

jenny-s51 (commented):
❗️ Updated the screenshots based on most recent updates.

src/components/credentials/credentials.js
```diff
@@ -125,7 +126,7 @@ class Credentials extends React.Component {
           rows={viewOptions.selectedItems.length}
           value={credentialsList}
         />
-      </Grid.Col>
+      </List>
```
cdcabrera, line 129: required
- lets remove the `List` component entirely since it's wrapping a form `textarea`. Plus, it doesn't appear to cause any visual difference by removing it

src/components/scanHostList/scanHostList.js
```diff
@@ -92,9 +91,9 @@ class ScanHostList extends React.Component {
 
     return (
       <div className="quipucords-infinite-results">
-        <Grid fluid onScroll={this.onScrollList} className="quipucords-infinite-list">
-          {hostsList.map(host => children({ host }))}
-        </Grid>
+        <List isPlain onScroll={this.onScrollList}>
+          {hostsList?.map(host => children({ host }))}
+        </List>
```
cdcabrera, line 96: required:
- looks like we might need both css classes on that div with the `onScroll` event, since one className applies width, and the other applies max-height
- and it looks like the consumers of this component don't apply `ListItem` so we'll have to apply it here too

```
        <div onScroll={this.onScrollList} className="quipucords-infinite-results quipucords-infinite-list">
          <List isPlain>
            {hostsList?.map(host => (
              <ListItem isPlain key={host.name}>
                {children({ host })}
              </ListItem>
            ))}
          </List>
        </div>
...
```


optional
- you'll probably have to leave the old css class since they're used elsewhere... create a `FixMe` note in `_view.scss` about consolidating both `quipucords-infinite-results` and `quipucords-infinite-list` into a single css class

src/components/credentials/credentialListItem.js
```diff
@@ -127,21 +128,15 @@ class CredentialListItem extends React.Component {
...
-          </Grid>
+          <List isPlain>
+            {item?.sources?.map(source => (
+              <Tooltip tooltip={dictionary[source.source_type]}>
+                <ListItem key={source.name} icon={<Icon type={typeIcon.type} name={typeIcon.name} />}>
+                  {source.name}
+                </ListItem>
+              </Tooltip>
+            ))}
+          </List>
```
cdcabrera, line 139: good effort on trying to get the tooltip in there! 

- unfortunately it jams a `span` around the `li` tag, 
- and it centers the tooltip on the element... which in this case has its width set by the largest list item so we'd also need to set the tooltip alignment to the left... 
 
...so lets avoid all of that

required:
- flip this back over to just wrapping the copy... 
   - not as cool, but maybe at some point if the tooltip gets upgraded to PF4 it won't wrap the `span` tag on there

```
        return (
          <List isPlain>
            {item?.sources?.map(source => (
              <ListItem key={source.name} icon={<Icon type={typeIcon.type} name={typeIcon.name} />}>
                <Tooltip tooltip={dictionary[source.source_type]}>{source.name}</Tooltip>
              </ListItem>
            ))}
          </List>
        );

**Review by cdcabrera — CHANGES_REQUESTED** [review]
Very close, and thank you for updating the test snapshots, they help with reviews!

File: src/components/credentials/credentials.js (hunk 3 above)
Comment (jenny-s51) [reply src/components/credentials/credentials.js]: Good call - updated!

File: src/components/scanHostList/scanHostList.js (hunk 4 above)
Comment (jenny-s51) [reply src/components/scanHostList/scanHostList.js]: Tested both of these options and it looks like we can set both classes on `List` as long as we remove that outer div!

File: src/components/credentials/credentialListItem.js (hunk 5 above)
Comment (jenny-s51) [reply src/components/credentials/credentialListItem.js:137]: Good point - updated!

**Comment by cdcabrera** [comment]
@jenny-s51 looks good! everything is working, nothing needs to change on the code side!

Go ahead and update the build output checks, after that we'll merge
```
$ yarn build
$ yarn test:integration-dev
```

cdcabrera (approved):

PR #153, opened 2022-09-01
feat(radio): add PF4 radio to forms

jenny-s51:
## What's included

Towards #104 : Radio component

### Notes

Emulating code style in https://github.com/RedHatInsights/curiosity-frontend/blob/main/src/components/form/checkbox.js

## Example

...
### Before
Screen Shot 2022-09-01 at 4 29 02 PM

### After
Screen Shot 2022-09-01 at 4 28 42 PM


cdcabrera (changes requested):
Good first attempt! 

Unfortunately interacting with the component and its internal state interferes with overall functionality. To recreate... Click through all 3 radio buttons

The radio buttons at some point require a double click to select the button, and then it starts losing track of state at the Redux level once you've managed to click all 3 radio buttons.

PR #170, opened 2022-09-22
fix(build): discovery-174 css font paths

cdcabrera:
#### Bug Fixes
1. **build** discovery-174 css font paths (6a4cf9e)

### Check the build
1. `$ yarn build`
1. confirm the `dist` output for the templates css file
   -  no longer contains references to `./assets/...`
   - but instead now uses `/static/media/...`

## Example
...

## Updates issue/story
[DISCOVERY-174](https://issues.redhat.com/browse/DISCOVERY-174)


bruno-fs (approved):
waaaay better to read than the regex I proposed on server side.

I tested building locally where it was breaking. LGTM :+1:

PR #192, opened 2022-11-07
Fix/api stage script

bruno-fs:
## What's included
Fix `yarn start:stage` script.

## How to test
Run `yarn start:stage` and validate all assets are loaded correctly.


bruno-fs:
It's tricky to have templates being generated automatically by webpack in sync. If the output from "dist" is not desired than the proposed changes won't help.

PR #201, opened 2023-04-11
Ansible controller integration

bruno-fs:
## What's included
Support for new "ansible controller/tower" scanner

## How to test
Ensure the workflow for adding/editing and deleting (when applicable) ansible credential/source and a scanner is working as expected.

### Local run check
2. `$ yarn build`
3. `$ yarn run api:stage` 

## Example
image

## Updates issue/story
[DISCOVERY-43](https://issues.redhat.com/browse/DISCOVERY-43)


cdcabrera:
Looks good so far @bruno-fs . I'll review more in-depth in a bit. At a quick glance this was the update for OpenShift
- https://github.com/quipucords/quipucords-ui/commit/1ac30164a170229d95bbacc01169b7674ad5b407#diff-c9e674b4dc402e05f627561a03fa45b3e001693235edd669a9df547cb160522b

We might need additional changes around `components/createCredentialDialog/createCredentialDialog.js` around lines `101` - `110`ish https://github.com/quipucords/quipucords-ui/blob/master/src/components/createCredentialDialog/createCredentialDialog.js#L101-L110. 

It's an additional switch statement that may need to be updated since it calls out the different `authorization types` passed on credential types

scripts/api.sh
```diff
@@ -206,15 +218,27 @@ stageApi()
...
   fi
 
   if [ ! "$UPDATE" = true ]; then
     startDB $DB_NAME
-
     if [ -z "$(docker ps | grep $NAME)" ]; then
       printf "\n"
       echo "Starting API..."
-      docker run -d --rm -p $PORT:443 -v $BUILD_DIR:$CLIENT_VOLUME -v $BUILD_DIR:$TEMPLATE_CLIENT_VOLUME -v $TEMPLATE_DIR:$TEMPLATE_REGISTRATION_VOLUME -e QPC_DBMS_HOST=$DB_NAME --link $DB_NAME:qpc-link --name $NAME $CONTAINER >/dev/null
+      docker run -d -p $PORT:443 \
```
cdcabrera, line 230: request:
... you drop the `--rm` back in the `docker run -d --rm -p $PORT:443` it appears to actually stop the Docker container, at least on Mac OS when you ctrl-c out of the terminal process

cdcabrera (changes requested):
First the positive...
I can confirm that the `JS` updates appear to function correctly and that the changes for
commit `feat(ansible-controller): add support for ansible data source` have my LGTM

Now the other...
After a bunch of trial and error, mostly error, and some Docker frustration... the end result is I was finally able to get the staging run to function correctly but with changes. I've noted them. 

Specifically the big issue is using a different image that isn't RHEL8 like Fedora off of Quay instead, which I noted.  If it helps with the Quay recommendation... originally the intent of the GUI staging run was to get a close approximation with live data so a 1 to 1 with production is only a loose requirement as long as we're getting "production like" data =)

bruno-fs:
Thanks for the thorough review, @cdcabrera!
I just pushed the suggested fixes. If they are OK I'll rebase/autosquash

PR #207, opened 2023-05-09
chore(deps): Updated decode-uri-component to ^0.3.0

abellotti:
chore(deps): Updated decode-uri-component to ^0.3.0

- Updated all indirect references to the decode-uri-component to be at least 0.3.0.

### DIFF
--- a/package.json
+++ b/package.json
@@ -154,7 +154,8 @@
     "yamljs": "^0.3.0"
   },
   "resolutions": {
-    "**/eslint": "7.19.0"
+    "**/eslint": "7.19.0",
+    "decode-uri-component": "^0.3.0"
   },
   "browserslist": {
     "production": [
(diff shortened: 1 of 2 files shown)


infinitewarp (approved):
I'm not a node/yarn expert, but LGTM! 👍

cdcabrera:
@abellotti couple of notes about the package updates, much appreciated though!

- First, yes, we could merge this and we still might, but we're going to aim to do a bulk update of all the NPMs short of Create React App. An issue should be being created by @nicolearagao 
- And, two, a subsequent set of build updates to move to webpack 5 needs to happen, working on that now ( an issue probably needs to be created). It will most likely nullify some existing working so heads up if you're burning some effort on the updates

nicolearagao:
> @abellotti couple of notes about the package updates, much appreciated though!
> 
> * First, yes, we could merge this and we still might, but we're going to aim to do a bulk update of all the NPMs short of Create React App. An issue should be being created by @nicolearagao
> * And, two, a subsequent set of build updates to move to webpack 5 needs to happen, working on that now ( an issue probably needs to be created). It will most likely nullify some existing work/effort, so heads up if you're burning some effort on the updates
> 
> And a sidenote...
> 
> > updating resolutions and yarn.lock directly creates long-term issues. The reason being if a package that requires it has built around some facet of the older package we're essentially creating a scenario for something breaking

@abellotti https://issues.redhat.com/browse/DISCOVERY-330 we are going to get that done after the delete-scan gets merged.

abellotti:
Thanks @cdcabrera and @nicolearagao for the notes. This makes sense.  Going to webpack 5 will also solve a couple of the outstanding older dependencies too. I will go ahead and close these 2 PRs since the whole lot will be taken care of via a bulk update.

PR #208, opened 2023-05-15
feature: add delete scan function

nicolearagao:
## What's included
fix(deleteScanDialog): locale strings
feature(deleteScan): adds scan deletion functionality

1. `$ yarn test`
2. confirm tests come back clean

### Local run check
1. `$ yarn start:stage`
1. add a
 - credential (any source)
 - source (same type of credential)
 - scan, confirm it
   - on scan window, a thrash icon is displayed according to style requirements to every added scan
   - when scaling the borwser down, it should populate the kebab menu
   - when selecting a scan to delete a confirmation message should be displayed
   - when deleting a scan, a confirmation message of the deletion should be displayed 

## Example
- Add credential
cred
- Add source
source
- Add scan
scan
- Visualize the trash icon
trash
- Confirmation if the item should be deleted
confirm-deletion
-Confirm item was deleted
deletion-confirmed

## Updates issue/story
feature(deleteScan): [discovery-76](https://issues.redhat.com/browse/DISCOVERY-76)


cdcabrera (changes requested):
Looks pretty good so far. Just a couple of minor mods need to happen to align with the storyboard. 

Attached some references for you and of course, as you noted, all those unit tests need to be updated.

src/components/scans/scansContext.js
```diff
@@ -28,6 +29,105 @@ const INITIAL_QUERY = {
...
+        {
+          context: ['heading', 'delete-scan'],
+          count: updatedScans.length,
+          name: updatedScans?.[0]?.[apiTypes.API_RESPONSE_SCAN_NAME]
+        },
+        [<strong />]
+      ),
+      body:
+        (updatedScans.length > 1 &&
+          updatedScans.map(({ [apiTypes.API_RESPONSE_SCAN_NAME]: name }) => name).join(', ')) ||
```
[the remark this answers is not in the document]
nicolearagao: I dont think it will ever be a priority to delete multiple scans at the same time so I will go with option 2

src/components/i18n/__tests__/__snapshots__/i18n.test.js.snap
```diff
@@ -1663,6 +1695,14 @@ Array [
...
     "key": "view.empty-state",
   },
+  Object {
+    "file": "./src/components/scans/scansContext.js",
+    "key": "form-dialog.confirmation",
+  },
+  Object {
+    "file": "./src/components/scans/scansContext.js",
+    "key": "form-dialog.confirmation",
+  },
```
infinitewarp, line 1705: This appears to be duplicated. Is that correct?

cdcabrera: Indeed, if you ref the file `src/components/scans/scansContext.js` and do a string search on the file you'll see `form-dialog.confirmation` pop up twice. 

The i18n snapshots @nicolearagao are a reference point for knowing where all keys are. It doesn't necessarily tell you if something is busted but like all JS snapshots in this repo its the first line of alert against "something changed, why did it change, did I mean to change that..." =) At some point I need to update the`CONTRIBUTING.md` with process how and whys.

src/redux/reducers/__tests__/scansReducer.test.js
```diff
@@ -78,7 +80,8 @@ describe('ScansReducer', () => {
       types.PAUSE_SCAN,
       types.RESTART_SCAN,
       types.START_SCAN,
-      types.GET_SCANS
+      types.GET_SCANS,
+      types.DELETE_SCAN
```
cdcabrera, line 84: required: around line `100` there is a `fulfilled` types check similar to `pending` and `error` just need to fill it in with the same `types.DELETE_SCAN` and update the related test snapshot. 

At a minimum it highlights that `DELETE_SCAN` is consistent in implementation with the other REST responses consumed by Redux

src/components/scans/scansTableCells.js
```diff
@@ -343,6 +347,22 @@ const actionsCell = ({
     }
   }
 
+  menuItems.push({
+    dropdownMenuItem: { title: t('table.label', { context: 'delete' }) },
```
cdcabrera, line 351: required: looks like we/I missed a small change that gets the small screensize menu to work... we need to update line `351` to have a `value` attribute like. It appears needed for when the Dropdown consumes the Array/list and generates.

```
dropdownMenuItem: { title: t('table.label', { context: 'delete' }), value: 'delete' },
```

cdcabrera (changes requested):

cdcabrera (approved):
lgtm! I can achieve a match and functionality by using both `$ yarn start` and `$ yarn start:stage`

I do recommend we have a double check from testing around deleting a scan that has just started. If that has already been confirmed props!

And finally, just need to work on the commit messaging a little. When we merge we can talk that over @nicolearagao

infinitewarp (approved):

PR #221, opened 2023-06-26
fix(createCredential): DISCOVERY-363 Use `ssh_passphrase`

mirekdlugosz:
Use `ssh_passphrase` field name, not `sshpassphrase`. The latter is rejected by server and prevents creation of credential.

## How to test

Try to create network credential with SSH passphrase against real Discovery / Quipucords server. Before this patch, server will reject such request as invalid.

## Updates issue/story
[DISCOVERY-363](https://issues.redhat.com/browse/DISCOVERY-363)


cdcabrera (approved):
Looks good! Legacy patch removed!

So this doesn't get lost, for the history. Unclear if this field ever work, or if the API at some point did consider it a single word, but as noted in the removed `ToDo`, carried over from the 2019 UI version work.

PR #224, opened 2023-06-30
docs: rename master branch to main

infinitewarp:
## What's included
Update a few outdated references for the old `master` branch to the new `main` branch.


nicolearagao (approved):

cdcabrera:
@infinitewarp We have this old Jenkins file here with a `master` branch ref... can we confirm this file can be deleted?
- https://github.com/quipucords/quipucords-ui/blob/main/Jenkinsfile

infinitewarp:
> @infinitewarp We have this old Jenkins file here with a `master` branch ref... can we confirm this file can be deleted?
> 
>     * https://github.com/quipucords/quipucords-ui/blob/main/Jenkinsfile

Good question! I have no idea what or if anything uses that file. 😕 I would be tempted to say, yes, let's delete that file now, and if we are surprised by something breaking, we can quickly add it back in another PR. Thoughts @cdcabrera ?

cdcabrera:
@infinitewarp from what I can see the new artifact just looks like a part of the github action scripting. I think the jenkins file used to be referenced from the quipucords repo... but I can't find that either... regardless it's using a `master` branch reference so it'll bust at some point if/when github's branch redirects stop from `master` -> `main`

lets remove it too!

cdcabrera (approved):
lgtm, let's add removing the jenkins file either here or as a separate commit

PR #235, opened 2023-08-11
refactor(testing): DISCOVERY-280 Add attributes to aid UI tests

mirekdlugosz:
## What's included

Add a bunch of `data-ouia-component-id` or `data-test` attributes on elements that we interact with in Camayoc UI tests.

### Notes

I don't know if these are _all_ elements we interact with, so there might be more PRs similar to this one later on.

Also, I think we need to tag a realase with this commit to include it in upstream container image? We might do it blind, or we might hold on merge and release until I have Camayoc PR ready.

## Updates issue/story

[DISCOVERY-280](https://issues.redhat.com/browse/DISCOVERY-280).


bruno-fs (approved):
looks ok to my unexperienced 👀

src/components/addSourceWizard/addSourceWizardStepOne.js
```diff
@@ -25,7 +25,12 @@ class AddSourceWizardStepOne extends React.Component {
       <FormState validateOnMount setValues={{ sourceType: type }} validate={this.isStepValid}>
         {({ values, handleOnEvent, handleOnSubmit }) => (
           <Form isHorizontal onSubmit={handleOnSubmit}>
-            <FormGroup role="radiogroup" isStack label={t('form-dialog.label', { context: 'source-type' })}>
+            <FormGroup
+              role="radiogroup"
+              fieldId="sourceType"
+              isStack
+              label={t('form-dialog.label', { context: 'source-type' })}
+            >
```
cdcabrera, line 33: It appears there may be a misunderstanding of the FormGroup components `fieldId` behavior. The current behavior was completed under #217. 

> The FormGroup component was designed to automagically apply the First Childs id consistently so a developer did not have to maintain 2 separate identifiers, which when dealing with form fields can become tedious and prone to errors.

There appears to be an unclear reason associated with this update. Why does this the `fieldId` for `addSourceWizardStepOne.js` have to change when there is already a generated ID that appears to be consistent?

src/components/createCredentialDialog/createCredentialDialog.js
```diff
@@ -226,6 +226,7 @@ const CreateCredentialDialog = ({
       {values[apiTypes.API_QUERY_TYPES.CREDENTIAL_TYPE] === 'network' && (
         <FormGroup label={t('form-dialog.label', { context: ['auth-type', 'create-credential'] })}>
           <DropdownSelect
+            ouiaId="auth_type"
```
cdcabrera, line 229: looks good!

optional: 
In the future consider modifying the `DropdownSelect` component to handle or apply the `ouiaId` when an `id` or `name` property is applied to `DropdownSelect`. It appears that functionality is missing currently.

The reason this thought popped up is that the `ouiaId` appears to be the same parameter name used within the form state

src/components/createCredentialDialog/createCredentialDialog.js
```diff
@@ -240,6 +241,7 @@ const CreateCredentialDialog = ({
           errorMessage={t('form-dialog.label', { context: ['username', 'error'] })}
         >
           <TextInput
+            ouiaId="username"
```
cdcabrera, line 244: required: Since `ouiaId` is the same as the name property modify `TextInput` to apply the `ouiaId` automatically

- https://github.com/quipucords/quipucords-ui/blob/main/src/components/form/textInput.js#L126

```
...
const updatedOuiaId = ouiaId || updatedName;

<PfTextInput
   id={updatedId}
   name={updatedName}
   ouiaId={updatedOuiaId}
```

OR you could even make a `data-test` attribute if you're looking for consistency against something else (I say stick with ouiaId)

```
// example code only end-result may require alteration
const updatedDataTest = props['data-test'] || updatedName;
```

Doing this should clean up most of the changes that are similar around TextInput.

cdcabrera (changes requested):
An optional, and 2 required changes. The required changes for `TextInput` and `TextArea` should turn most of the current updates into "automagic". 
> The caveat to this is the snake case that's currently being used gets converted to whatever the name attribute happens to be... which appears as a combination of snake and lower camel.

No obligation to use, but we added a couple of commits in my fork that can get you started, feel free to squash them into your work.
- https://github.com/cdcabrera/quipucords-ui/commits/add-test-locators

src/components/viewToolbar/viewToolbarSelect.js
```diff
@@ -106,6 +106,7 @@ const ViewToolbarSelect = ({
       position={position}
       selectedOptions={selectedOption}
       data-test={`toolbarSelect_${filter}`}
+      ouiaId={`toolbarSelect_${filter}`}
```
cdcabrera, line 109: question: unclear why both of these `data-test` and `ouiaId` are being used, are both necessary?

src/components/viewToolbar/viewToolbarFieldSortButton.js
```diff
@@ -63,6 +63,7 @@ const ViewToolbarFieldSortButton = ({ t, useOnClick: useAliasOnClick, useQuery:
         onClick={() => !isEmpty && onClick({ value: updatedDirection })}
         variant={ButtonVariant.plain}
         data-test="toolbarSortButton"
+        ouiaId="toolbar_sort_button"
```
cdcabrera, line 66: question: same question, are both `data-test` and `ouiaId` required?

src/components/viewToolbar/viewToolbarFieldSort.js
```diff
@@ -74,6 +74,7 @@ const ViewToolbarFieldSort = ({ t, useOnSelect: useAliasOnSelect, useView: useAl
         onSelect={onSelectUpdated}
         selectedOptions={selectedOption?.replace(/^-/, '')}
         data-test="toolbarSortType"
+        ouiaId="toolbar_sort_type"
```
cdcabrera, line 77: question: same question, are both `data-test` and `ouiaId` required?

src/components/viewToolbar/viewToolbarTextInput.js
```diff
@@ -155,6 +155,7 @@ const ViewToolbarTextInput = ({
         value={currentValue}
         placeholder={t('toolbar.label', { context: ['placeholder', 'filter', filter] })}
         data-test="toolbarSearchName"
+        ouiaId="toolbar_search_name"
```
cdcabrera, line 158: question: same question, are both test attributes required?

src/components/createScanDialog/createScanDialog.js
```diff
@@ -343,6 +345,7 @@ class CreateScanDialog extends React.Component {
           <TextArea
             isDisabled={!checked.jbossEap && !checked.jbossFuse && !checked.jbossWs && !checked.jbossBrms}
             name="displayScanDirectories"
+            ouiaId="display_scan_directories"
```
cdcabrera, line 348: required: Apply the same automagic of `name` and autogenerated `ouiaId` for `TextInput` to `TextArea`

src/components/addSourceWizard/addSourceWizardStepOne.js
mirekdlugosz: I was looking at [PatternFly docs](http://v4-archive.patternfly.org/v4/components/form/#formgroup), which claim that `fieldId` is "required when a role of "group" or "radiogroup" is passed in." That's why added it, as I noticed it's not specified and generated one is the same as value of first item (which is strange, if harmless).

Now I see that most / all components here are custom components that are thin wrappers around PatternFly components. So we can skip `fieldId`, but one will be generated before we use PatternFly component.

Removed in ae106e3f01300fe89c58b0c8cf1ef74a584fb4cc . I'm adding commits to make review easier, they should be all squashed before / during merge.

src/components/createCredentialDialog/createCredentialDialog.js
mirekdlugosz: I'm not sure about this. Our generated ids are not globally unique, and as a result, `data-ouia-component-id` starts to be duplicated in snapshots. This is probably less of a problem in app itself, where we pay more attention to use globally unique `name`s.

Also, while `ouiaId` might be the same as `name` in many cases, I **want** to keep them decoupled. I intend to use mostly `data-ouia-component-id` in Camayoc, and I don't want to be forced to update Camayoc immediately every time names change in UI. I don't yet know if this is a viable approach, but I would like to give it a try.

src/components/viewToolbar/viewToolbarSelect.js
mirekdlugosz: This is mostly for consistency on Camayoc side. Specifying `ouiaId` on pretty much all interactive elements we care about allows us to use it consistently in automation, and probably introduce a number of helpers and heuristics (i.e. "if not specified otherwise, assume string is ouia-id"). When some elements have `data-test` and some have `ouia-id`, it only increases effort needed to maintain automation on Camayoc side.

src/components/createCredentialDialog/createCredentialDialog.js
cdcabrera: Given the implementation of the code, decoupled still works with the update where `ouiaId` is the generated "fallback" and overridden by whatever custom ouiaID you input. Its the best of both, and appears to also cover cases where overlooked fields also exist

src/components/createScanDialog/createScanDialog.js
cdcabrera: Good catch! I think as long as your testing framework can read it `naive` works. It's a functional update that can be expanded or relaxed if/when the next version of Patternfly is implemented

src/components/createCredentialDialog/createCredentialDialog.js
mirekdlugosz: Added in 0bcb6d08763d75a83d68f73ce10cc10e079e597f , along with changes to TextArea, discussed below.

I don't like how we lose unique ouia ids in snapshots, but as long as CI is happy, this is not a hill to die on.

src/components/form/textArea.js
```diff
@@ -160,7 +165,8 @@ TextArea.defaultProps = {
   onClear: helpers.noop,
   onKeyUp: helpers.noop,
   resizeOrientation: TextAreResizeOrientation.default,
-  value: ''
+  value: '',
+  ouiaId: ''
```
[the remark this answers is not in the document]
mirekdlugosz: Done in 0bcb6d08763d75a83d68f73ce10cc10e079e597f for both `input` and `textarea`. I _hope_ they are pretty much the same now, at least that is my intent.

cdcabrera (approved):
lgtm!

If you wanted to help future you, or ui dev out, updating this section in [`CONTRIBUTING.md`](https://github.com/quipucords/quipucords-ui/blob/main/CONTRIBUTING.md#reserved-css-classnames-and-attributes) about why/when we use `data-test` vs `ouiaId` that'd be helpful.

src/components/createCredentialDialog/createCredentialDialog.js
mirekdlugosz: I meant before this PR, `data-ouia-component-id` in snapshot was globally unique. Now you can notice some repetitions, when multiple components in the same snapshot have `generatedid-` value.

Obviously, I am not aware of some larger architecture choices that impact this result, and how it changed historically.

[ouia spec says ids must be unique withing surrounding context](https://ouia.readthedocs.io/en/latest/README.html#ouia-component), not globally unique (like standard HTML `id` attribute), and I guess our in snapshot still are. And even if not, it doesn't really impact anything.

All in all, this is just something that gets under my skin and what I personally dislike. Since you are OK with this change, and there is no actual technical problem with it, I think it's best to move forward with what we have in that regard.

mirekdlugosz:
@cdcabrera added one final change to attributes, 758483da48adfb0a10fd5698c2ec4187a03b9d2f . I noticed sometimes tests failed to add source, when adding source is particularly slow (pending state is displayed longer than a split second), and automation works especially fast. Automation wants to close a dialog, but when source is being added, this displays "are you sure you want to quit?" popup instead of closing.

The obvious solution is for automation to wait for dialog to reach stable state, but there's no good element for us to wait for. So I exposed state in `data-test` attribute.

I'll add changes to contributing later, and if you are fine with above change, I can squash commits and we can merge.

src/components/addSourceWizard/addSourceWizardStepThree.js
```diff
@@ -19,7 +19,10 @@ import { EMPTY_CONTEXT, translate } from '../i18n/i18n';
  * @returns {React.ReactNode}
  */
 const AddSourceWizardStepThree = ({ add, error, fulfilled, pending, name, t }) => (
-  <EmptyState className="quipucords-empty-state">
+  <EmptyState
+    className="quipucords-empty-state"
+    data-test-state={(error && 'error') || (pending && 'pending') || (fulfilled && 'fulfilled') }
+  >
```
cdcabrera, line 25: ...looks like this code base could benefit from the data-test IDs applied through the translation tool

**fix:**
the linter is asking you to remove the space between the fulfilled condition and bracket 
```
(fulfilled && 'fulfilled')[space]}
```
**OR recommendation:**
what you have works, but...
Did you also try leveraging the existing conditions?

Starting around line `26` you could look at applying a `data-test` to each of the `EmptyStateIcon` components. You'd have to confirm, but on a light check it does appear `fulfilled` is coming through on the SVG, and `ouiaId` does kinda.

Another way around it is to repeat what you did may have done before and just apply a `<span data-test="[state]">` around a wrapping element.




```
{error && <EmptyStateIcon data-test="error" icon={ExclamationCircleIcon} color={red.value} />}
{fulfilled && <EmptyStateIcon data-test="fulfilled" icon={OutlinedCheckCircleIcon} color={green.value} />}
{pending && <EmptyStateIcon data-test="pending" icon={Spinner} />}
```

Screen Shot 2023-08-21 at 10 11 22 AM

cdcabrera (changes requested):

src/components/addSourceWizard/addSourceWizardStepThree.js
mirekdlugosz: I went straight to this first solution I had, didn't try anything else :) .

I tried adding attribute to icon now, and it works. Works like a charm on Camayoc side, too. Done that in 3bfc641f4261b005e1aea307d110d86aabd8600f

cdcabrera (approved):

src/components/createCredentialDialog/createCredentialDialog.js
cdcabrera: Ah, so "generatedid-" is for testing snapshots only... we removed the random hash that gets appended otherwise the snapshots would be updating on every test

ruda:
I ran locally (Fedora 38) `yarn test` on this MR.

```
Test Suites: 104 passed, 104 total
Tests:       358 passed, 358 total
Snapshots:   546 passed, 546 total
Time:        41.663 s
Ran all test suites matching /.\/src/i.
Done in 51.32s.
```

All good.

There is a warning message regarding `isReadOnly` element, if you think it is relevant to take, here is it.

```
PASS src/components/form/__tests__/checkbox.test.js
  ● Console

    console.error
      Warning: React does not recognize the `isReadOnly` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `isreadonly` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
```

ruda (approved):
LGTM!

PR #237, opened 2023-08-24
scripts/api.sh with podman and docker support

ruda:
## What's included

* Add support to detect and execute `api.sh` script with podman or docker. In the past, there was only support for docker execution. Related to `yarn start` execution.

### DIFF
--- a/scripts/api.sh
+++ b/scripts/api.sh
@@ -319,7 +319,7 @@ stopApi()
 
   QPC_REPO="https://github.com/quipucords/quipucords.git"
   QPC_IMAGE_CONTAINER="quay.io/quipucords/quipucords:latest"
-  MOCK_IMAGE_CONTAINER="palo/swagger-api-mock"
+  MOCK_IMAGE_CONTAINER="docker.io/palo/swagger-api-mock"
   DATADIR="$(pwd)/.qpc"
   DATADIR_REPO="$(pwd)/.qpc/quipucords"
   DATADIR_REPO_QPC="$(pwd)/.qpc/quipucords/quipucords"
--- a/scripts/api.sh
+++ b/scripts/api.sh
@@ -14,12 +14,12 @@ checkContainerRunning()
   while [ $COUNT -le $DURATION ]; do
     sleep $DELAY
     (( COUNT++ ))
-    if [ -z "$(docker ps | grep $CHECKONE)" ]; then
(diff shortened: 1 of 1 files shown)


scripts/api.sh
```diff
@@ -349,16 +350,30 @@ stopApi()
...
+  fi
   if [ -z "$(docker -v)" ]; then
-    printf "\n${RED}Docker missing, confirm installation and running.${NOCOLOR}\n"
+    printf "\n${YELLOW}Docker missing, confirm installation and running.${NOCOLOR}\n"
+  else
+    printf "\n${GREEN}Found docker CLI.\n"
+    DOCKER="docker"
+  fi
+  if [ -z "$DOCKER" ]; then
+    printf "\n${RED}Docker or podman missing, giving up.${NOCOLOR}\n"
```
cdcabrera, line 366: You can probably be more transparent instead of saying "giving up"

```
printf "\n${RED}Missing container tooling. Unable to set DOCKER. Docker, or podman, is required.${NOCOLOR}\n"
```

cdcabrera (changes requested):
looks good! 

On the initial conditional checks I think we can simplify the messaging, see my prior comment. 

A nice to have, or future piece, might be to expand the [`CONTRIBUTING.md`](https://github.com/quipucords/quipucords-ui/blob/main/CONTRIBUTING.md#install-tooling), looks like we don't really go into depth on `podman`

cdcabrera:
@ruda we're going to be removing `dev` to help avoid confusion in the future. After main is rebased you'll need to rebase from `main` and point this PR in that direction

cdcabrera:
as of 20230830 rebased

scripts/api.sh
ruda: I'm going take it all what you've suggested.

infinitewarp (approved):

ruda (commented):
The suggestions for improvements are now there. The documentation can be improved later, if needed (which I don't think so now).

mirekdlugosz (approved):
I checked out this branch, ran `yarn start` and browser opened UI with autogenerated data. This is the only thing I ever used containers for in this repo, so these changes look good enough for me :+1:

cdcabrera:
Will rebase and merge, thank you @mirekdlugosz !

cdcabrera (approved):

PR #243, opened 2023-09-25
chore(refact): Disable the scan Merge Reports functionality.

abellotti:
The functionality has been broken for a while in the UI, and needs to be updated to leverage the async version of the API. For now we won't expose the Merge report Button and leave this as a CLI only functionality until the UI can be properly updated.

Current feature:

Screenshot 2023-09-25 at 5 32 39 PM

Screenshot 2023-09-25 at 5 32 56 PM

After the disabling the feature (Merge reports button is hidden):

Screenshot 2023-09-25 at 5 31 27 PM

### DIFF
--- a/src/components/scans/__tests__/__snapshots__/scans.test.js.snap
+++ b/src/components/scans/__tests__/__snapshots__/scans.test.js.snap
@@ -34,6 +34,11 @@ exports[`Scans Component should handle multiple display states, pending, error, 
             <Button
               isDisabled={true}
               onClick={[Function]}
+              style={
+                {
+                  "display": "none",
+                }
+              }
               variant="primary"
             >
               t(table.label, {"context":"merge-reports"})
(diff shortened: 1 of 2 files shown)


bruno-fs (commented):
LGTM, but let's wait @cdcabrera 👀

abellotti:
@cdcabrera when you get a chance please.

cdcabrera (changes requested):
This technically works! I think it's good as a temporary solution.

I would recommend a note in file referencing the Jira issue. If there isn't a Jira issue, I recommend you create one so there's some tracking on this, things seem to get lost.

Also recommend you

- include that Jira story reference in the commit message
- Refactor is an actual type you can use for the commit message...  message could be `refactor(scans): dis-[issue number] disable merge reports` (putting the entire DISCOVERY-[ISSUE NUM] in there is a little tedious)

abellotti:
There you go @cdcabrera 

Thank you for lending your 👀

cdcabrera (approved):

PR #251, opened 2023-10-11
feat(acs): add support for acs data source

nicolearagao:
## What's included
Support for new ACS scanner

## How to test
Ensure the workflow for adding/editing and deleting (when applicable) acs credential/source and a scanner is working as expected.

### Local run check
2. `$ yarn build`
3. `$ yarn run api:stage` 

## Updates issue/story
https://issues.redhat.com/browse/DISCOVERY-350


src/components/contextIcon/contextIcon.js
```diff
@@ -126,6 +127,8 @@ const ContextIcon = ({ symbol, ...props }) => {
       return <PficonNetworkRangeIcon {...props} />;
     case ContextIconVariant.openshift:
       return <OpenshiftIcon {...props} />;
+    case ContextIconVariant.acs:
+      return <OpenshiftIcon {...props} />;
```
bruno-fs, line 131: I wonder if there's a better icon for ACS

abellotti: we probably want the same icon we see when we install ACS from OperatorHub in OpenShift.

image

src/components/addCredentialType/__tests__/__snapshots__/addCredentialType.test.js.snap
```diff
@@ -10,6 +10,10 @@ exports[`ToolbarFieldGranularity Component should export select options: fieldOp
     "title": [Function],
     "value": "openshift",
   },
+  {
+    "title": [Function],
+    "value": "acs",
```
mirekdlugosz, line 15: I don't know why (might be backend thing), but when I do Credentials -> Add Credential -> ACS credential, then "Source type" in new dialog window is "acs", all lowercase. I think it should be all uppercase, "ACS". All other types are correctly capitalized, like "Network" or "OpenShift".

image

public/locales/en.json
```diff
@@ -114,6 +115,8 @@
     "label_option_vcenter_credential": "vCenter Credential",
     "label_option_ansible": "Ansible",
     "label_option_ansible_credential": "Ansible Credential",
+    "label_option_acs": "ACS",
+    "label_option_acs_credential": "ACS credential",
```
mirekdlugosz, line 119: This should be capitalized, for consistency with other labels around:

```suggestion
    "label_option_acs_credential": "ACS Credential",
```

src/components/addSourceWizard/__tests__/__snapshots__/addSourceWizardStepOne.test.js.snap
```diff
@@ -86,6 +86,27 @@ exports[`AddSourceWizardStepOne Component should render a basic component: basic
...
+            id="sourceType-acs"
+            name="sourceType"
+            type="radio"
+            value="acs"
+          />
+          <label
+            class="pf-c-radio__label"
+            for="sourceType-acs"
+          >
+            t(form-dialog.label, {"context":"acs"})
```
mirekdlugosz, line 98: Same here - I don't know why, but "ACS" is all lowercase.

image

mirekdlugosz (dismissed):

src/components/addCredentialType/__tests__/__snapshots__/addCredentialType.test.js.snap
cdcabrera: Around `line 60` in `en.json` if you add the key with context under `form-dialog`
```
"label_acs": "ACS",
```
... this will display the human intended display of "ACS" @nicolearagao . This should also apply to the Source and Credential displays.

The reason
You just happened to miss one string =). Because of the way "context" works on strings you're currently displaying the fallback context passed to the translate function... if you look at @bruno-fs old PR you'll catch the similar `Ansible` update, https://github.com/quipucords/quipucords-ui/pull/201/files#diff-1ef07a459de08da33c9f950f32e4e739683a016d219c16503bf8c3a5e2c973c7R173

cdcabrera (commented):
@nicolearagao is there an intended update to [API spec](https://github.com/quipucords/quipucords/blob/main/docs/swagger.yml) for `ACS`?

src/components/contextIcon/contextIcon.js
nicolearagao: yes, that would be ideal.

nicolearagao: @cdcabrera I see that on patternfly the icon for ACS is not available. https://www.patternfly.org/2022.05/guidelines/icons
Is there an easy way to import a product icon to patternfly or is it not the way it works ? Thanks!
https://docs.google.com/presentation/d/1M1DhoaO3Rnkv5IKWSSZnZvH0SOxsbVDHyDz0k2SgzTY/edit#slide=id.g21c2e5b05dd_0_0

src/components/addCredentialType/__tests__/__snapshots__/addCredentialType.test.js.snap
cdcabrera: Appears to now be working as @mirekdlugosz  noted

src/components/addSourceWizard/__tests__/__snapshots__/addSourceWizardStepOne.test.js.snap
cdcabrera: This appears to be resolved too!

cdcabrera (approved):
lgtm
- aligns to the similar update for ansible data source
-  `$ yarn start:stage` (shortcut for the above steps noted) and `$ yarn start` appear to function as intended
   - adding credentials
   - adding sources
   - adding scans 

I'll defer to others in case I overlooked anything

bruno-fs (approved):

PR #270, opened 2023-12-19
chore(lint): Fix Prettier and ESLint configs and fix errors

mturley:
There were conflicting prettier rules in the `.eslintrc` file and `.prettierrc` files. The linter was disagreeing with my editor and giving me errors after running prettier on save. It turns out `.eslintrc.js` was being used and `.eslintrc` wasn't, and the rules inside `.eslintrc.js` matched the existing code most closely. So I deleted the redundant configs.

I also fixed how the prettier config/plugin is configured in `.eslintrc.js` (see their [recommended configuration](https://github.com/prettier/eslint-plugin-prettier#recommended-configuration)), then ran `yarn format` and `yarn lint:fix` (a new script I added that runs `eslint --fix`).

After this, there were some remaining eslint errors I needed to fix by hand.

Note that I only fixed lint errors, there are still some warnings present:

```
/Users/mturley/git/quipucords-ui/src/components/TypeaheadCheckboxes.tsx
   14:8  warning  Using exported name 'TimesIcon' as identifier for default export                                                                                                                                                                                                                       import/no-named-as-default
   71:6  warning  React Hook React.useEffect has missing dependencies: 'isOpen' and 'options'. Either include them or remove the dependency array                                                                                                                                                        react-hooks/exhaustive-deps
  152:6  warning  React Hook React.useEffect has a missing dependency: 'placeholder'. Either include it or remove the dependency array. If 'setActivePlaceholder' needs the current value of 'placeholder', you can also switch to useReducer instead of useState and read 'placeholder' in the reducer  react-hooks/exhaustive-deps

/Users/mturley/git/quipucords-ui/src/pages/sources/AddSourceModal.tsx
  57:6  warning  React Hook React.useEffect has a missing dependency: 'typeValue'. Either include it or remove the dependency array  react-hooks/exhaustive-deps

/Users/mturley/git/quipucords-ui/src/pages/sources/SourcesListView.tsx
  211:6  warning  React Hook React.useEffect has a missing dependency: 'columnOrderMap'. Either include it or remove the dependency array  react-hooks/exhaustive-deps

✖ 5 problems (0 errors, 5 warnings)
```


mturley:
Oops, didn't mean to open this to main

mturley:
There we go.

bruno-fs:
Just to be sure I'm following what's happening here: you removed a conflicting config in `.eslintrc` and all the other changes are the result of running `yarn format`?

mturley:
> Just to be sure I'm following what's happening here: you removed a conflicting config in `.eslintrc` and all the other changes are the result of running `yarn format`?

That's correct. This was based on a Slack discussion with @gitdallas, and was mostly a selfish change because my editor was confused. I'd be fine with not making this change if it's too disruptive.

Also hey 👋 I'm Mike, I'm helping out a little with upgrading the usage of the react-table-batteries dependency that I'm working on

cdcabrera:
The issue is Patternfly seed was layered on top of the original build...

Prettier was originally accessed inside the eslint config… if you look here you can see the settings on the original main file… https://github.com/quipucords/quipucords-ui/blob/main/.eslintrc#L125-L132

You just need to pick a path @mturley ... pick the path of separating Eslint and Prettier configuration and removing one of those configs OR combine Eslint and Pretter like it is currently in `main`

mturley:
I see. Looks like I picked the wrong path, I'll switch back over to what was in the eslint files. Maybe delete the redundant one? looks like on main there's only `.eslintrc`, idk where the `.eslintrc.js` came from (I guess PF seed?). Which should be used?

mturley:
To get CI to pass it looks like I may need to also fix some other lint errors.

cdcabrera:
@mturley dealers choice. Eslint is updating their stance on configuration files and the naming of said files, basically it'll be changing at some point anyway... if this was me I'd take the shortest/most simplistic path

mturley:
It looks like the `.eslintrc.js` is what was actually being used, as running it with the `.eslintrc` is missing dependencies like the `airbnb` config package. I'll delete that file and go with `.eslintrc.js`, then fix lint errors that are coming up.

mturley:
@bruno-fs @cdcabrera @gitdallas I think it should be good to go now. I updated the description above with the new changes to fix eslint.

cdcabrera (changes requested):

package.json
```diff
@@ -36,7 +36,8 @@
     "test:coverage": "jest --coverage",
     "eslint": "eslint --ext .tsx,.js ./src/",
```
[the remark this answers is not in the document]
mturley: @cdcabrera good catch, fixed that. I also had to add some eslint-ignore comments in src/common/helpers.ts because there's a ton of `any` usage in there and I think maybe fixing that is outside the scope of this PR

mturley:
Ugh hang on, now Prettier and ESLint are disagreeing again. 1 sec.

mturley:
Ok, it should be ready now I think.

mturley:
Looks like CI didn't like my commit messages, squashed to one commit and added the scope prefix

__mocks__/fileMock.js
```diff
@@ -1 +1 @@
-module.exports = 'test-file-stub';
\ No newline at end of file
+module.exports = "test-file-stub";
```
cdcabrera, line 1: The double quote changes in all build files is a remanent of your prior changes to eslint... if you're not going to include the ability to alter these files in the NPM `eslint` script you may want to revert all single to double quote updates

mturley: ah, I thought I had fixed that, sorry. I guess we should only bother with linting stuff in `src/` anyway

mturley: Thanks @cdcabrera. Pushed a version that reverts style changes to all files outside `./src`, so I'm only changing things affected by the lint config changes. These were indeed remnants of when I was using the prettier CLI directly.

mturley:
@cdcabrera do you mind approving the CI workflows? I don't have permissions to re-run those

cdcabrera (approved):
Functionality appears to include prettier, and passes on my setup.

Be good to have someone else double check

@mturley the commit linter is saying your git message is over 65 characters at 70 =)

mturley:
> the commit linter is saying your git message is over 65 characters at 70
lol, of course. should be fixed now @cdcabrera

gitdallas (approved):

PR #289, opened 2024-01-09
feat(source): add delete button

nicolearagao:
Add delete button for source without functionality.


src/pages/sources/SourcesListView.tsx
```diff
@@ -261,6 +261,15 @@ const SourcesListView: React.FunctionComponent = () => {
       .finally(() => setPendingDeleteSource(undefined));
   };
 
+  const hasSelectedSources = () => {
+    return Object.values(selectedItems).filter(val => val !== null).length > 0;
```
gitdallas, line 265: trying to figure out what all of this is. shouldn't `selectedItems` already be an array and you could just do `selectedItems?.length`?

nicolearagao: well, I didnt know better. I will push mods and ask for re-review

gitdallas (approved):

PR #302, opened 2024-01-22
Refactor/api service

nicolearagao:


src/hooks/api/useScanApi.ts
```diff
@@ -0,0 +1,55 @@
...
+      addAlert,
+      queryClient
+    );
+  };
+
+  // const onDeleteSelectedItems = () => {
+  //   // add logic
+  //   const selectedItems = [];
+  //   console.log('Deleting selected credentials:', selectedItems);
+  // };
```
infinitewarp, line 45: Dead code?

infinitewarp: Or placeholder for future code?

nicolearagao: placeholder for future code

src/app/routes.tsx
```diff
@@ -59,7 +59,7 @@ const AppRoutes = (): React.ReactElement => {
   const nav = useNavigate();
   const location = useLocation();
 
-  axios.get('https://0.0.0.0:9443/api/v1/users/current/').catch(() => {
+  apiService.get('https://0.0.0.0:9443/api/v1/users/current/').catch(() => {
```
infinitewarp, line 62: Is the longer-term intent to eventually create functions in the `apiService` module and to move these long call arguments (like `get('https://0.0.0.0:9443/api/v1/users/current/')`) into them to abstract away their implementatino? In this case, something like `getCurrentUser`? Otherwise, I don't understand the benefit of simply replacing `axios` with `apiService`.

nicolearagao: yes, that was the plan. I abstracted it too much and too early and broke the application to a point I didn't know what was wrong, so I had to rollback on changes and retrace my steps. I am going to need @gitdallas to move forward with this, this refactor was way way out of my comfort zone. Any feedback is definetely welcome!

infinitewarp (approved):
As someone who doesn't write a lot of React/TypeScript, I long for module/class/function docstrings to explain briefly their purpose, but I understand that a lot of this is probably just boilerplate React/TS code and the purpose is very obvious to seasoned developers. 🤷 

In the Python world, we encourage and enforce at least minimal inline documentation with `ruff` (and previously via `flake8-docstrings`), and I know some other mainstream languages have similar encouragement/enforcement tools, but I don't know about JS/TS. Are there standard or common tools we could use here also? As we want to broaden and democratize the ownership and maintenance of `quipucords-ui`, I would like to see more inline documentation so that contributors don't have to be (as much) experts in the existing code base.

(I'm not blocking this PR to require docs, but I wanted to put these thoughts somewhere to start a conversation.)

gitdallas (changes requested):
i know we did our video code review, but some reminders along with some other things i've noticed going through

src/app/routes.tsx
nicolearagao: rolled back on those changes, will open separate pr with new hook.

src/components/sessionContext/SessionProvider.tsx
```diff
@@ -49,7 +49,7 @@ const SessionProvider: React.FC<{ children: React.ReactNode }> = ({ children })
     queryKey: [WHO_AM_I_QUERY],
     refetchOnWindowFocus: !helpers.DEV_MODE,
     queryFn: () => {
-      return axios.get(process.env.REACT_APP_USER_SERVICE_CURRENT || '').then(res => {
+      return apiService.get(process.env.REACT_APP_USER_SERVICE_CURRENT || '').then(res => {
```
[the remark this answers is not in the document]
nicolearagao: rolled back on those changes, will open separate pr with those hooks.

src/hooks/api/useCredentialApi.ts
```diff
@@ -0,0 +1,40 @@
+import { API_CREDS_LIST_QUERY } from 'src/constants/apiConstants';
+import apiHandler from 'src/services/apiUtils';
+import { CredentialType } from 'src/types';
+import apiService from '../../services/apiService';
+import useApiConfig from '../apiService/apiConfig';
+import usePendingState from '../common/usePendingState';
+
+const useCredentialApi = () => {
+  const { pendingItem: pendingDeleteCredential, setPendingItem: setPendingDeleteCredential } =
```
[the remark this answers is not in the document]
nicolearagao: Can I keep it or would you prefer if I changed it ?

src/hooks/api/useCredentialApi.ts
```diff
@@ -0,0 +1,40 @@
...
+    usePendingState<CredentialType>();
+  const selectedItems = [];
```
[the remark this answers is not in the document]
nicolearagao: I will remove this because this is just a placeholder for a future implementation.

src/hooks/api/useCredentialApi.ts
```diff
@@ -0,0 +1,40 @@
...
+      `Credential "${credential.name}" deleted successfully`,
+      `Error removing credential ${credential.name}.`,
+      [API_CREDS_LIST_QUERY],
+      addAlert,
+      queryClient
+    ).finally(() => setPendingDeleteCredential(undefined));
+  };
+
+  const onDeleteSelectedCredentials = () => {
+    const itemsToDelete = Object.values(selectedItems).filter(val => val !== null);
```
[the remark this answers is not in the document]
nicolearagao: check comment above, this part should be implemented in other pr.

src/hooks/api/useCredentialApi.ts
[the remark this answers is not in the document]
gitdallas: keep

src/hooks/alertsHelper.ts
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
```
gitdallas, line 1: file should be useAlerts though

src/hooks/api/useScanApi.ts
```diff
@@ -0,0 +1,56 @@
...
+      successMessage: `Scan "${scan.id}" deleted successfully`,
+      errorMessage: `Error removing scan ${scan.id}.`,
+      queryKey: [API_SCANS_LIST_QUERY],
+      addAlert: addAlert,
+      queryClient: queryClient
+    }).finally(() => setPendingDeleteScan(undefined));
+  };
+
+  const onRunScan = (scan: ScanType, payload) => {
+    apiHandler<any>({
```
gitdallas, line 29: can we add a type for some if they don't match any we have?

src/hooks/common/usePendingState.ts
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+
+const usePendingState = <T>() => {
+  const [pendingItem, setPendingItem] = React.useState<T>();
```
gitdallas, line 4: wonder if this should be an array, so we can have pendingItems (which may be only one item in some instances, like editing a source... but could also work for when we are selecting several items, if that makes sense)

src/services/apiUtils.ts
```diff
@@ -0,0 +1,40 @@
+import { getUniqueId } from '@patternfly/react-core';
+import { QueryClient } from '@tanstack/react-query';
+
+interface ApiHandlerProps<T> {
+  apiServiceMethod: Function;
+  url: string;
+  payload: T;
+  successMessage: string;
+  errorMessage: string | null;
```
gitdallas, line 9: would rather see undefined than null, which can also be done like
```suggestion
  errorMessage?: string;
```

src/hooks/api/useScanApi.ts
nicolearagao: This code will be modified soon when we add scans features.

src/hooks/common/usePendingState.ts
nicolearagao: I like the idea, but this change will break the code in a million parts. Let's add this in another pr.

src/services/apiUtils.ts
```diff
@@ -0,0 +1,40 @@
+import { getUniqueId } from '@patternfly/react-core';
+import { QueryClient } from '@tanstack/react-query';
+
+interface ApiHandlerProps<T> {
+  apiServiceMethod: Function;
```
gitdallas, line 5: for here and line 11, don't use `Function` as a type as it won't provide any type safety. here you'll probably want something like `(url: string, payload: any) => Promise<AxiosResponse<any, any>>`, for example

gitdallas: (with changes proposed in chat, this becomes obsolete)

src/pages/scans/ScansListView.tsx
```diff
@@ -27,37 +34,86 @@ import {
...
-  const removeAlert = (key: React.Key) => {
-    setAlerts(prevAlerts => [...prevAlerts.filter(alert => alert.key !== key)]);
+  const onRunScan = (payload: ScanType) => {
+    /**
+     * Initiates a scan for the provided source, handles success, error, and cleanup operations.
+     *
+     * @param {ScanType} payload - The payload containing source information to start the scan.
+     */
+    console.log('run scan:', payload);
+    runScan(payload);
```
gitdallas, line 103: the runScan rest call actually takes a list of IDs, as such we'll need to pass the id in here instead of the entire ScanType object. as such, you should probably change `payload` name to `scan` as it isn't the payload that is getting sent.. then do `scan.id` or whatever.

src/services/apiConfig.ts
```diff
@@ -0,0 +1,18 @@
+/**
```
gitdallas, line 1: i don't believe this is being used anymore

gitdallas (changes requested):

src/hooks/api/useCredentialApi.ts
```diff
@@ -0,0 +1,44 @@
...
+ */
+import React from 'react';
+import axios from 'axios';
+import { CredentialType } from 'src/types';
+
+const useCredentialApi = () => {
+  const [pendingDeleteCredential, setPendingDeleteCredential] = React.useState<CredentialType>();
+
+  const deleteCredential = (credential?: CredentialType) => {
+    /**
```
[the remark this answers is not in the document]
gitdallas: FYI: this is one comment for many instances

gitdallas (approved):

PR #310, opened 2024-02-05
feat(Scans): implement scans modal with dl

gitdallas:
included some commented out checkbox progress for multi dl (for whenever that gets implemented in the rest api)


src/hooks/api/useScanApi.ts
```diff
@@ -45,14 +45,31 @@ const useScanApi = () => {
...
+    return axios.get(`${process.env.REACT_APP_SCANS_SERVICE}${scanId}/jobs/`);
+  }
+
+  const downloadReport = (reportId: number) => {
+    return axios({
+      url: `${process.env.REACT_APP_REPORTS_SERVICE}${reportId}/`,
+      method: 'GET',
+      responseType: 'blob',
+    });
+  }
```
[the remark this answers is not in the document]
gitdallas: if i just do a normal get request, it will not download the file

nicolearagao (approved):
LGTM
Before merging, we gotta fix the error causing yarn audit to fail.

gitdallas:
@nicolearagao - it looks like the audit is complaining about npm-check-updates which is a dev dep. we probably don't need this as a dep as you can always run npm-check-updates with npx, and we've been using dependabot anyway... but i'm not sure why it's triggering for just this pr, that dep was prior to this

gitdallas:
it looks like the security issue with ncu was discovered between the last pr and this. updating ncu removed the audit issues. again we could probably remove ncu if you want at some point as it is just something to run locally (rather than an actual project dependency) so you can always run via npx (which just temporary gets the package).

src/pages/scans/components/ScansModal.tsx
```diff
@@ -0,0 +1,140 @@
...
+    //     setSelectedJobs((prevSelected) => {
+    //         const otherSelected = prevSelected.filter((j) => j !== job.id);
+    //         return isSelecting && isJobSelectable(job) ? [...otherSelected, job.id] : otherSelected;
+    //       });
+    // };
+    // const areAllJobsSelected = false;
+
+    // const selectAllJobs = (isSelecting: boolean = true) => {
+    //     setSelectedJobs(isSelecting ? selectableJobs.map((j) => j.id) : []);
+    // };
```
infinitewarp, line 62: Why do we have lots of commented-out code here and elsewhere in this function?

gitdallas: from the original comment on this pr: "included some commented out checkbox progress for multi dl (for whenever that gets implemented in the rest api)"

infinitewarp (approved):

PR #312, opened 2024-02-06
fix(credentials): DISCOVERY-534 Don't try to update read-only fields

mirekdlugosz:
## What's included

Fix for DISCOVERY-534 - when editing a credential, don't send back `created_at` and `updated_at` fields. They are managed by backend.


gitdallas (approved):

infinitewarp (changes requested):
Please amend commit message to include this line:

```
Relates to JIRA: DISCOVERY-534
```

…to make the Red Hat JIRA bot happy. 🤷‍♂️ 

edit: Also, I wouldn't worry about the failing yarn audit check. We should merge this despite that failure because, at least for the short term, we already know and expect the audit to fail on the main branch.

nicolearagao (approved):

PR #331, opened 2024-03-20
Fix/minor issues

nicolearagao:
Minor fixes to mainly address disparities between current implementation and UX sketches.
- refactor order of tables
- fix(source): re-order fields to match UX sketches
- fix(sources): avoid inconsistencies in sources
- fix(scans): use correct icon when no scans available
- fix(sources): correct no sources available text

### DIFF
--- a/src/pages/sources/components/AddSourceModal.tsx
+++ b/src/pages/sources/components/AddSourceModal.tsx
@@ -124,6 +124,13 @@ const AddSourceModal: React.FC<AddSourceModalProps> = ({ source, type, onClose, 
                 }}
               />
             </FormGroup>
+            <FormGroup label="Credentials" fieldId="credentials" isRequired>
+              <TypeaheadCheckboxes
+                onChange={setCredentials}
+                options={credOptions}
+                selectedOptions={credentials?.map(String) || []}
+              />
+            </FormGroup>
             {isNetwork ? (
               <>
                 <FormGroup label="Search addresses" isRequired fieldId="hosts">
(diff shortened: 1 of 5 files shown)


src/pages/sources/components/AddSourceModal.tsx
```diff
@@ -79,7 +79,7 @@ const AddSourceModal: React.FC<AddSourceModalProps> = ({ source, type, onClose,
       credentials: credentials.map(c => Number(c)),
       hosts: values['hosts'].split(','),
       name: values['name'],
-      port: !isNetwork ? '443' : values['port'] || '22',
+      port: !isNetwork ? values['port'] || '443' : values['port'] || '22',
```
gitdallas, line 82: so values can always override, even if not `isNetwork`?

if so, probably better to do like... `values['port'] || (isNetwork ? '22' : '443')`

gitdallas (approved):
looks good, one nit i commented on

src/pages/sources/components/AddSourceModal.tsx
```diff
@@ -155,24 +162,32 @@ const AddSourceModal: React.FC<AddSourceModalProps> = ({ source, type, onClose,
...
+                    value={getValue('port')}
+                    placeholder="Optional"
+                    type="text"
+                    id="source-port"
+                    name="port"
+                    onChange={ev => {
+                      setValue('port', (ev.target as HTMLInputElement).value);
+                    }}
+                  />
+                  <HelperText>Default port is 443</HelperText>
```
ruda, line 187: For what source type this 443 (HTTPS) refers to? I mean, for network (SSH) the default port is 22.

nicolearagao: yes, network default port is specified somewhere else.

nicolearagao: values['port'] || (isNetwork ? '22' : '443')

src/pages/sources/components/AddSourceModal.tsx
nicolearagao: fixed

PR #333, opened 2024-04-01
build(Containerfile): initial support for building as standalone container

bruno-fs:
## What's included

...

## Example

...


.env
```diff
@@ -46,3 +46,4 @@ REACT_APP_USER_SERVICE_CURRENT=/api/v1/users/current/
 REACT_APP_USER_SERVICE_LOGOUT=/api/v1/users/logout/
 
 REACT_APP_STATUS_SERVICE=/api/v1/status/
+REACT_APP_AUTH_TOKEN_SERVICE=/api/v1/token/
```
infinitewarp, line 49: It seems like there's a missing newline at the end of this line/file (indicated by the the red ⛔  symbol in github).

Not a big deal, but it'll make future commits and diffs cleaner since most editors include the newline.

infinitewarp (approved):
This looks reasonable to me! Pretty standard copy and build commands and boilerplate nginx config.

nicolearagao (approved):

.env
mirekdlugosz, line 49: As far as I can tell, all these variables are listed in `CONTRIBUTING.md`, so I guess this new one should be added there, too.

bruno-fs: updated!

bruno-fs: Thanks for pointing that out. Just updated `CONTRIBUTING.md` with this info.

cdcabrera, line 49: required:
if this is replacing an actively used path you'll also need to update the
`.env.development` file with the same `param` but using the original path for development

cdcabrera (changes requested):
I'll assume the container work is solid since its being used elsewhere, so lgtm!

The dotenv work just needs a little bit of an expansion to work with both
- prod
- & development
See comment above

.env
bruno-fs: this is required. V1 UI didn't need it because login was done through the django template view for login.

updated `.env.development` accordingly, thanks for pointing this out =D

infinitewarp (approved):

mirekdlugosz (approved):

cdcabrera (approved):
Looks good! Nothing blocking from the dotenv side, go for it

Future note... technically all dotenv parameters are strings wrapping a dotenv param into a template like 

```
`${process.env.[some-param]}`
```

could just be
```
process.env.[some-param]
```
This is generally why we end up moving them to a centralized file to help refine, or cast, the values to their final state (one side-effect of using the template/ticks is "undefined" becomes a string called "undefined" =) ... welcome to JS!

PR #336, opened 2024-04-08
fix: ds-625 lowerCamelCase all files in src dir

nicolearagao:
## What's included
Ensure that for consistency all files are cased the same. lowerCamelCase is going to be the UI's goto agnostic preference.

## Updates issue/story

[DISCOVERY-625](https://issues.redhat.com/browse/DISCOVERY-625)


mirekdlugosz (approved):
The changes look sane. I am not in position to say if there is anything this PR is missing, but even if there is, that can always be fixed later.

I see there's [eslint rule for camelcase variable names](https://eslint.org/docs/latest/rules/camelcase) and [eslint plugin that provides rule for file names](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/filename-case.md). Perhaps it would be good idea to include them, so any future deviations are caught automatically?

cdcabrera (approved):
lgtm!

PR #337, opened 2024-04-09
build: ds-626 reactivate eslint for webpack

cdcabrera:
## What's included

- build: ds-626 reactivate eslint for webpack

### Notes

- reactivates eslint on webpack startup for both "terminal" and "browser console" output

### Local run check
1. `$ yarn start`
1. you should see something similar to the following message.
   - all notifications are non-blocking beyond glaring notification
   - development can still continue
   - all fixes will be handle in subsequent Jira/stories/issues
Screenshot 2024-04-09 at 1 50 44 PM

## Updates issue/story

[DISCOVERY-626](https://issues.redhat.com/browse/DISCOVERY-626)

### DIFF
--- a/config/webpack.dev.js
+++ b/config/webpack.dev.js
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/no-var-requires */
-// const ESLintPlugin = require('eslint-webpack-plugin');
+const ESLintPlugin = require('eslint-webpack-plugin');
 const MiniCssExtractPlugin = require('mini-css-extract-plugin');
 const { merge } = require('webpack-merge');
 const { setupWebpackDotenvFilesForEnv, setupDotenvFilesForEnv } = require('./build.dotenv');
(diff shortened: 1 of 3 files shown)


config/webpack.dev.js
```diff
@@ -20,11 +20,12 @@ module.exports = merge(
       ...setupWebpackDotenvFilesForEnv({
         directory: RELATIVE_DIRNAME,
         env: MODE
+      }),
```
nicolearagao, line 23: @cdcabrera are we going to worry about cache later ? I know that we want to fix the whole project right now, but do we want to lint only changed files and skip the rest in the near future?

nicolearagao (approved):
LGTM

config/webpack.dev.js
cdcabrera: @nicolearagao Yes, we do have to migrate the build over to NPM and a few other tidbits to help resolve oddities. 

We're going with this as a low-hanging fruit. It still has to be done so while we're developing (or when we get back to that point) we're receiving linting feedback, see [ds-629](https://issues.redhat.com/browse/DISCOVERY-629)

PR #343, opened 2024-04-23
feat(deploy): ds-358 add support for enabling V1 vs V2 UI

abellotti:
- Add support for enabling V1 vs V2 UI
- Add support for user-specified server certificates
- Make sure /var/log/nginx is created before starting nginx


abellotti:
recreated 344

PR #345, opened 2024-04-25
1 of 4, build: ds-663 update local run, staging run for podman

cdcabrera:
## What's included

- build: ds-663 update local run, staging run

### Notes
- Moves to podman and the newer container setup
   - **There is a temporary caveat** as the v1 work moves over from Django templates (if it ever does)... the developer still needs to login. In order to login the script `$ yarn start:stage` attempts to open 2 paths to the server
      - one for login around `https://localhost:9443/login/`
      - and another for development around `http://localhost:3000/client/sources`
- This is prep for all subsequent build updates and verification they are working

### Local run check
**This run no longer requires a container**
1. `$ yarn start`
1. the browser will attempt to open around  `http://localhost:3000/client`
1. confirm you are able to access the randomized API response version of the UI

### Local run check staging
1. Confirm podman or podman desktop is up and running
1. `$ yarn start:stage`
1. the browser will attempt to open around  `http://localhost:3000/client` AND `https://localhost:9443/login/`
   - You will need to login FIRST using the path `https://localhost:9443/login/` before attempting to use `http://localhost:3000/client` ... this is the limitation while we move away from Django templates for v1
1. confirm you are able to access the UI and it behaves as expected from the related Quipucords:latest image
   - Note: because v1 does NOT use the same auth method aspects of v2 the API will potentially throw a `4XX` level error, this is expected during the potential move to a "login for v1"

## Example

...

## Updates issue/story

[DISCOVERY-663](https://issues.redhat.com/browse/DISCOVERY-663)


cdcabrera:
As noted by @infinitewarp I'll remove ALL references to the `D-word`
@infinitewarp @mirekdlugosz @ruda @bruno-fs @nicolearagao @abellotti

scripts/apiStage.sh
```diff
@@ -0,0 +1,136 @@
...
+    $PODMAN run -itd --rm \
+      -e QPC_SERVER_PASSWORD=$PASSWORD \
+      -e QPC_DBMS=sqlite \
+      -p $PORT:443 \
+      -v "${HOME}"/.local/share/discovery/log/:/var/log \
+      -v "${HOME}"/.local/share/discovery/data/:/var/data \
+      -v "${HOME}"/.local/share/discovery/sshkeys/:/sshkeys \
+      --tls-verify=false \
+      --name $NAME \
+      $CONTAINER
```
infinitewarp, line 82: I haven't tested this in _all_ environments, but `OSTYPE` is a standard variable set by bash and bash-like shells, and I confirmed it is set with the expected value in macOS 14.2 and RHEL9. I'm pretty sure this will work for the foreseeable future.

```suggestion
    MOUNT_ARGS=$([[ "$OSTYPE" == "darwin"* ]] && echo "" || echo ":z")

    $PODMAN run -itd --rm \
      -e QPC_SERVER_PASSWORD=$PASSWORD \
      -e QPC_DBMS=sqlite \
      -p $PORT:443 \
      -v "${HOME}"/.local/share/discovery/log/:/var/log"$MOUNT_ARGS" \
      -v "${HOME}"/.local/share/discovery/data/:/var/data"$MOUNT_ARGS" \
      -v "${HOME}"/.local/share/discovery/sshkeys/:/sshkeys"$MOUNT_ARGS" \
      --tls-verify=false \
      --name $NAME \
      $CONTAINER
```

config/webpack.dev.js
```diff
@@ -52,7 +52,14 @@ module.exports = merge(
       },
       watchFiles: {
         paths: ['src/**/*', 'public/**/*']
-      }
+      },
+      proxy: [
+        {
+          context: ['/credentials', '/jobs', '/ping', '/reports', '/sources', '/scans', '/status', '/token', '/users'],
+          target: 'http://0.0.0.0:8000',
```
infinitewarp, line 59: Should this be `127.0.0.1`?

config/webpack.stage.js
```diff
@@ -37,25 +38,33 @@ module.exports = merge(
...
         directory: DIST_DIR
       },
       watchFiles: {
         paths: ['src/**/*', 'public/**/*']
-      }
+      },
+      proxy: [
+        {
+          context: ['/api'],
+          target: 'https://0.0.0.0:9443',
```
infinitewarp, line 64: Should this be `127.0.0.1`?

config/webpack.dev.js
cdcabrera: So kinda... [these are pointed at what was happening in v2 as part of the container allowance, very similar to the dotenv settings in v2](https://github.com/quipucords/quipucords-ui/blob/v2-alpha/.env.development#L6)... webpack itself needs to be pointed at 127.0.0.1. This one I have to sort-of wait to confirm our Linux user can access it

config/webpack.stage.js
cdcabrera: same response as above, the actual location of the update happened in the dotenv file `.env` with an update to `HOST=127.0.0.1`

config/webpack.dev.js
cdcabrera: Looks like a confirmation we can skip this update

config/webpack.stage.js
cdcabrera: Same as above, the dotenv file update appears to have been enough

ruda (approved):

package.json
```diff
@@ -27,13 +27,11 @@
...
   "scripts": {
-    "api:build": "run-s -l build:pre build:js build:post",
-    "api:dev": "sh ./scripts/api.sh -p 5000 -t dev",
-    "api:specs": "sh ./scripts/api.sh -t specs",
-    "api:review": "sh ./scripts/api.sh -p 9443 -t review",
-    "api:stage": "sh ./scripts/api.sh -p 5001 -t stage",
-    "api:stop": "sh ./scripts/api.sh -t stopApi",
-    "api:update": "npm run api:build; sh ./scripts/api.sh -t update",
+    "api:dev": "node ./scripts/apiDev.js --port 8000 --file https://raw.githubusercontent.com/quipucords/quipucords/master/docs/swagger.yml",
+    "api:spec": "open https://petstore.swagger.io/?url=https://raw.githubusercontent.com/quipucords/quipucords/master/docs/swagger.yml",
```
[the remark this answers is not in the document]
cdcabrera: I have a solution for this one too... appears to work locally, i'll update accordingly
```
"api:spec": "SPEC=https://petstore.swagger.io/?url=https://raw.githubusercontent.com/quipucords/quipucords/master/docs/swagger.yml; xdg-open $SPEC || open $SPEC",
```

package.json
```diff
@@ -48,12 +46,10 @@
     "build:template-css": "run-s build:template-scss build:template-css-font",
     "release": "changelog",
     "start": "run-p -l api:dev start:js",
-    "start:specs": "open https://petstore.swagger.io/?url=https://raw.githubusercontent.com/quipucords/quipucords/master/docs/swagger.yml",
     "start:js": "export NODE_ENV=development; webpack serve --config ./config/webpack.dev.js",
     "start:js-stage": "export NODE_ENV=development; webpack serve --config ./config/webpack.stage.js",
     "start:dev": "npm start",
-    "start:review": "open https://127.0.0.1:9443/login/; run-s build && run-s -l api:review;",
-    "start:stage": "open https://localhost:5001/login/; run-s api:build && run-p -l api:stage start:js-stage",
+    "start:stage": "open https://localhost:9443/login/; run-p -l api:stage start:js-stage",
```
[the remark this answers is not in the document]
cdcabrera: Moved the open over to the sh script now and applied `xdg-open` || 'open'

scripts/apiDev.js
```diff
@@ -0,0 +1,126 @@
+const { join } = require('path');
```
nicolearagao, line 1: Do u think error handling here for riskier operations that might fail would be overkill @cdcabrera ?

scripts/apiStage.sh
```diff
@@ -0,0 +1,138 @@
+#!/usr/bin/env bash
```
nicolearagao, line 1: 'set -euo pipefail' might help catch errors early and executing subsequent commands when something fails. 

https://explainshell.com/explain?cmd=set%20-euxo%20pipefail

scripts/apiStage.sh
```diff
@@ -0,0 +1,138 @@
...
+  $PODMAN system prune -f
+  printf "${GREEN}$PODMAN cleaning success.${NOCOLOR}\n"
+}
+#
+#
+# Stop all QPC containers
+#
+stopApi()
+{
+  $PODMAN stop -t 0 $($PODMAN ps --filter name="qpc*")
```
nicolearagao, line 49: The command below treats the entire output as one block, which helps keep the grouping of container names or IDs intact, especially important if they include spaces.

`$PODMAN stop -t 0 "$($PODMAN ps --filter name='qpc*')"`

nicolearagao (approved):
LGTM

scripts/apiDev.js
cdcabrera: In the related NodeJS scripting... We will say, good error handling is the foundation to quick debugging. 

An explanation to the script
1. You may be referencing the console logging to help call out missing args/params from the script... Overkill we can sort of see in the sense
   - ...that all of this work is a one-time patch where no-one will ever change this script again... 
   - But it may not be overkill if this work is applicable to the v2 branch, where we may modify some of this work.
1. We wrap the the related  `curl` in a try/catch. If that's the case, overkill that's debatable
   - ... yes it could be overkill, but in this case we used it to help relate error color coding in the console output to help sighted users have error logging that stands out from the other logging

If you're referencing something else we'll need more detail

nicolearagao: Thanks for the comprehensive answer

bruno-fs (approved):

PR #348, opened 2024-04-26
style: ds-653 Organize src files into standard dir

nicolearagao:
What's included
ds-653 UI: Code style - Organize src files and directories into standard directories
 
Notes
Changes applied according to reorg files/dirs [doc](https://docs.google.com/document/d/1nvzasTxg2sKP1YHkjh9fFqvemoW6lTn_RU9qSdkKNjM/edit#heading=h.j77k5mtuyj6)

How to test
2. `$ npm test`

2. `$ npm run test:dev`

### Local run check
2. `$ npm run start:stage`

### Check the build
2. `$ npm run build`

Updates issue/story
[DISCOVERY-653](https://issues.redhat.com/browse/DISCOVERY-653)


cdcabrera (changes requested):
LGTM! Very close!

I can get it to run with podman, get the existing tests to pass, and the build to output (and the newish local run which we'll apply in the near future).

However, and this parts on me, looks like I missed the nested `hooks/api` files. One last set of moves and we should be good...

Move the hooks underneath `src/hooks/api` so they're right under the `src/hooks`. Reason, we've already defined the hooks do by using `api` in the file name, the directory with the same name just adds redundancy.

That would look like

```
src
  hooks
    useCredentialsApi.ts
    useScanApi.ts
    useSourceApi.ts
```

src/components/refreshTimeButton/refreshTimeButton.tsx
```diff
@@ -7,7 +7,7 @@ import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { Button } from '@patternfly/react-core';
 import { RebootingIcon } from '@patternfly/react-icons';
-import { helpers } from '../../common/helpers';
+import { helpers } from 'src/helpers/helpers';
```
cdcabrera, line 10: lets be consistent here, you used relative paths in some of the other components for `helpers` so this line should be
```
import { helpers } from '../../helpers';
```

src/helpers/__tests__/helpers.test.ts
```diff
@@ -4,7 +4,7 @@
  * Validates accuracy and consistency of information display helpers. Includes various input scenarios and formatting checks.
  */
 import moment from 'moment';
-import { CredentialType } from 'src/types';
+import { CredentialType } from 'src/types/types';
```
cdcabrera, line 7: relative path consistency
```
import { CredentialType } from '../../types/types';
```

src/helpers/helpers.ts
```diff
@@ -8,7 +8,7 @@ import { PropertyPath } from 'lodash';
 import _get from 'lodash/get';
 import _set from 'lodash/set';
 import moment from 'moment';
-import { CredentialType } from 'src/types';
+import { CredentialType } from 'src/types/types';
```
cdcabrera, line 11: relative path consistency
```
import { CredentialType } from '../../types/types';
```

src/helpers/queryHelpers.ts
```diff
@@ -7,7 +7,7 @@
 import { TableState } from '@mturley-latest/react-table-batteries';
 import { useQuery } from '@tanstack/react-query';
 import axios from 'axios';
-import helpers from 'src/common';
+import helpers from 'src/helpers';
```
cdcabrera, line 10: relative path consistency
```
import helpers from './helpers';
```

Just a heads up. We haven't been calling this out but we have consistency with named vs default export linting piece coming up. We're going to be moving everything to "named" exports instead of "default" exports... so eslint after you do this may complain about not using this format just follow what the linter tells you... this is one of the many linting things we'll be correcting long term

```
import { helpers } from './helpers';
```

cdcabrera (changes requested):
Every works... but while we're in here... one less thing to fix next round when we start doing test updates

A few more changes for consistency... moving the files helped expose this. 

If I missed any unit tests that aren't named the same as the related files, please go ahead and update those names as well

cdcabrera (approved):
base ask of reorg files is done, lgtm!

PR #349, opened 2024-04-26
4 of 4, build(deps): ds-661 pf5 package updates

cdcabrera:
## What's included

- build(deps): ds-661 pf5 package updates

### Notes
- This update WILL alter any tests that are dependent on CSS selectors @mirekdlugosz @ruda 
- Apply pf5 components while attempting to retain the custom styling for v1, like font-size
   - The reason we did this was to avoid having to do a massive overhaul of the v1 styling, easier to slightly modify than rewrite 
   - The about modal will be the most "dramatic" alteration. The original image that the Patternfly 4 component moved location, we opted to apply the same background image used on login in v1 and now also in v2, see below screenshots
- ~this may inadvertently solve [DISCOVERY-162](https://issues.redhat.com/browse/DISCOVERY-162) the truncating Credentials listing when add/editing a source~
   - there was a separate change at the API level that bumped the default paging to `100`. The new "truncate" for credentials in the list is 100 entires, see issue for comment updates

1. `$ yarn test`

### Local run check
1. `$ yarn start:stage`
1. confirm that behavior is still working, as originally was (including any bugs, bugs will be fixed in subsequent PR work) for
   - viewing Source, Scans, Credentials
   - adding Sources, Scans, Credentials

### Check the build
1. `$ yarn build`

## Example

~Indirect solution for [DISCOVERY-162](https://issues.redhat.com/browse/DISCOVERY-162)~

Screenshot 2024-05-01 at 4 48 36 PM
Screenshot 2024-05-01 at 4 48 45 PM
Screenshot 2024-05-01 at 4 48 55 PM
Screenshot 2024-05-01 at 4 49 08 PM

## Updates issue/story

[DISCOVERY-661](https://issues.redhat.com/browse/DISCOVERY-661)


infinitewarp (approved):

nicolearagao (approved):
LGTM

mirekdlugosz (changes requested):
This introduces a problem with add source wizard. In step 1 I am supposed to select a source type. But clicking values does not change anything - default "Network Range" is still selected. I don't see any error in console either.

It looks like `handleOnEvent` does not modify `values.sourceType` (and possibly doesn't do other things it is supposed to be doing). I wasn't able to find this function actual code, so I can't verify that. From what I see, none of `addSourceWizardStepOne`, `formState`, `formGroup` or `Radio` were modified in this commit, so I have no idea why this would suddenly stop working.

mirekdlugosz (approved):
Yes, now the issue is fixed :) 

I think this is safe to merge, but we can't tag the new release with these changes until Camayoc PR is opened.

PR #350, opened 2024-04-26
refactor: ds-390 remove pause, restart scans

cdcabrera:
## What's included

- refactor: ds-390 remove pause, restart scans

### Notes
- removes v1s pause, restart scans functionality
-  ~the icons contain a "restart" icon that displays for `start` in the event a user has a failed scan. we can replace this if needed with just the "play" icon, see attached screenshot~
   - the only remaining icons are `play` and `stop` see the attached screenshot 

### Coverage and basic unit test check
1. update the NPM packages with `$ yarn`
1. `$ yarn test`
1. confirm tests come back clean

### Local run check
1. update the NPM packages with `$ yarn`
1. `$ yarn start:stage`
1. confirm the functionality that remains for scans is for
   ~_*note: the icons contain a "restart" icon that displays for `start` in the event a user has a failed scan_~
  _*note: the only remaining icons are now `play` and `stop`_
   - `cancel`/`stop`
   - `start`/`play`

### Check the build
1. update the NPM packages with `$ yarn`
1. `$ yarn build`
1. confirm tests come back clean

## Example

Screenshot 2024-05-06 at 2 12 56 PM

## Updates issue/story

[DISCOVERY-390](https://issues.redhat.com/browse/DISCOVERY-390)
blocked by #345 #346 #347 #349

PR #365, opened 2024-05-06
fix(scans): ds-98 disable expandable sources

cdcabrera:
## What's included

- fix(scans): ds-98 disable expandable sources

### Notes
- the sources displayed under scans suffers from an sequential id issue, this bypasses the issue by disabling the expansion field
   - we discussed using the same display v2 uses but until it can be determined v2 is pulling the correct information we're taking the direct approach of disabling the field all together 
      - the number of sources that are displayed is an aspect of what v2 uses to show sources
-  we deprecated the old "scanSourceList.js" component as opposed to deleting it simply because it was faster to type the word vs deleting the file, test file, test snapshots, etc. if someone wants us to remove it entirely we can also do that

### Local run check
1. `$ yarn start:stage`
1. confirm that the sources displayed on scans is now disabled, no longer clickable

## Example

Screenshot 2024-05-06 at 2 31 05 PM

## Updates issue/story

[DISCOVERY-98](https://issues.redhat.com/browse/DISCOVERY-98)
blocked by, requires #350

### DIFF
--- a/src/components/scans/__tests__/__snapshots__/scans.test.js.snap
+++ b/src/components/scans/__tests__/__snapshots__/scans.test.js.snap
@@ -181,8 +181,6 @@ exports[`Scans Component should handle multiple display states, pending, error, 
                     t(table.label_status, {"context":"cell","count":0}, [object Object],[object Object])
                   </Tooltip>,
                   "dataLabel": "t(table.header, {"context":"sources"})",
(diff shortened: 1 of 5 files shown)


infinitewarp (approved):

nicolearagao (approved):
LGTM

bruno-fs (approved):

PR #366, opened 2024-05-06
chore: drop support for brms

infinitewarp:
## What's included
https://issues.redhat.com/browse/DISCOVERY-674

As discussed in Slack ([link](https://redhat-internal.slack.com/archives/C02QSNF1UKE/p1715008099070109?thread_ts=1715007693.427309&cid=C02QSNF1UKE)), we can drop JBoss BRMS support immediately.

This change should merge and release around the same time as https://github.com/quipucords/quipucords/pull/2644 and https://github.com/quipucords/qpc/pull/329

I ran `yarn test` and it appears to pass.
```
Test Suites: 104 passed, 104 total
Tests:       359 passed, 359 total
Snapshots:   549 passed, 549 total
Time:        44.446 s
Ran all test suites matching /.\/src/i.
✨  Done in 78.34s.
```


cdcabrera:
this goes in after the other v1 updates
#350 #365 #367

cdcabrera (approved):
lgtm! 

confirmed in both local runs that the checkbox value is no longer available for user access, and removal appears thorough

only thing we might recommend is... this is probably more along the lines of a `refactor` rather than a `chore` since we're altering logic to do this update

...something like
`refactor(createScanDialog): ds-674 drop support for brms`

PR #367, opened 2024-05-06
fix(credentials): ds-205 disable edit for encrypted fields

cdcabrera:
## What's included

- fix(credentials): ds-205 disable edit for encrypted fields

### Notes
- Moves the credential "update" service from `put` to `patch`
- We opted for the most simplistic solution on this one... 
   - Technically the all asterisk values are still returning from the API as before, it's just NOW, before submission of the updated cred the values are removed
- Disables edits for the following encrypted fields. If additional fields are required we can add them in...
   - auth_token
   - ssh_passphrase
   - become_password
   - password 

### Local run check
1. make sure `podman` is running
1. `$ yarn start:stage`
1. add a credential with an encrypted field type
   - after adding the credential, edit it, submit it and confirm the field has NOT been updated with `*`asterisk characters . You will most likely need access to the same tests @ruda performed to determine this. We used form submission on edited credentials as our verification 

## Example

...

## Updates issue/story

[DISCOVERY-205](https://issues.redhat.com/browse/DISCOVERY-205)
blocked by #350 #365

### DIFF
--- a/src/components/createCredentialDialog/createCredentialDialog.js
+++ b/src/components/createCredentialDialog/createCredentialDialog.js
@@ -166,6 +166,10 @@ const CreateCredentialDialog = ({
 
     // clean for submitting an edited cred
     if (edit) {
+      delete updatedValues[apiTypes.API_QUERY_TYPES.AUTH_TOKEN];
+      delete updatedValues[apiTypes.API_QUERY_TYPES.SSH_PASSPHRASE];
+      delete updatedValues[apiTypes.API_QUERY_TYPES.BECOME_PASSWORD];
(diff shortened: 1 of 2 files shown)


bruno-fs (approved):

infinitewarp (changes requested):
Commenting here to block merging because I think this may have broke something. I ran against a real local version of quipucords, and updating what I believe is a previously-working credential broke it. I'll retrace my steps to verify and return with some details.

infinitewarp (approved):
Disregard my previous comment. The host I was targeting had changed since the last time I used it, and I didn't catch that in time. Trying again with newly verified host and credentials, this change does appear to work as advertised.

PR #369, opened 2024-05-14
fix(credentials): ds-681 hide token in network

mirekdlugosz:
Don't display "Token" authentication type for Network credential. Not only it doesn't make any sense, such credential can't be created anyway.

Relates to JIRA: DISCOVERY-681

### DIFF
--- a/src/components/createCredentialDialog/createCredentialDialog.js
+++ b/src/components/createCredentialDialog/createCredentialDialog.js
@@ -30,27 +30,6 @@ import { translate } from '../i18n/i18n';
 const becomeMethodTypeOptions = ['sudo', 'su', 'pbrun', 'pfexec', 'doas', 'dzdo', 'ksu', 'runas'];
 
 /**
- * Generate authentication type options.
- *
- * @type {{title: Function | React.ReactNode, value: string, selected: boolean}[]}
- */
-const authenticationTypeOptions = [
-  {
-    title: () => translate('form-dialog.label', { context: ['option', 'sshKey'] }),
-    value: 'sshKey'
-  },
-  {
-    title: () => translate('form-dialog.label', { context: ['option', 'token'] }),
-    value: 'token'
-  },
-  {
-    title: () => translate('form-dialog.label', { context: ['option', 'usernamePassword'] }),
-    value: 'usernamePassword',
-    selected: true
-  }
-];
-
-/**
  * ToDo: updating a creds password should immediately reset the entire field
  * When a user attempts to update a cred the password displayed is not decrypted. Existing behavior has the
  * appearance that any user who updates the field but leaves parts of the old password actually submit a combination
@@ -64,7 +43,6 @@ const authenticationTypeOptions = [
  * @fires onSubmit
  * @fires onValidateForm
  * @param {object} props
- * @param {Array} props.authenticationOptions
  * @param {Array} props.becomeMethodOptions
  * @param {Function} props.t
  * @param {Function} props.useCredential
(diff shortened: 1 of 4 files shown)


src/components/createCredentialDialog/createCredentialDialog.js
```diff
@@ -86,6 +86,7 @@ const CreateCredentialDialog = ({
   const submitCredential = useAliasOnSubmitCredential();
   const isShhKeyfile = credential?.[apiTypes.API_QUERY_TYPES.SSH_KEYFILE] && true;
   const isToken = credential?.[apiTypes.API_QUERY_TYPES.AUTH_TOKEN] && true;
+  const networkAuthenticationOptions = authenticationOptions.filter(item => item.value !== 'token');
```
cdcabrera, line 89: question:
this does work! 

...but if it doesn't exist why don't we remove the entire option from lines `42` -`45` instead and avoid renaming and filtering all together, it's less logic? Added bonus, the list of options doesn't appear to be used anywhere except this file and the related unit test file

- https://github.com/quipucords/quipucords-ui/blob/main/src/components/createCredentialDialog/createCredentialDialog.js#L42-L45

If the concern is "[we may get it working again in the future]" I think we should feel pretty good about your commit message with the Jira issue number called out as a reference

cdcabrera (changes requested):
Technically the solution works, but curious about an alternative

src/components/createCredentialDialog/createCredentialDialog.js
mirekdlugosz: My main concern is that we have single create/edit component for all credential types, and token **is** valid credential type - for OpenShift and RHACS. It's only not valid for network (and many others, but they don't give the option to switch authentication type).

So instead of risking breaking OpenShift / RHACS, or doing extensive testing they are not broken, I decided to play it safe and do the change that I know shouldn't affect anything else. For context, that was added back in commit 1ac30164a170229d95bbacc01169b7674ad5b407, which does claim it being done for OpenShift.

If you are confident we can just remove the option from lines 42-45, I can do it this way.

cdcabrera: Thank you for clarifying

So if `token` is still valid for `OpenShift` and `RHACS`, just not `network`, then we need to adjust the work a little bit more. It appears the other cred types display a standalone `token` field. 

The original/current setup is confusing. Unclear why we picked having the options in standalone list/array but it probably had to do with some aspect of being reusable at some point, except we never reused it

We should be good since what you're adjusting only applies to `network`, hence your rename is spot on, `networkAuthenticationOptions`.

Path one... the faster option
1. Remove the "token" option between lines `42-45`
2. rename that component prop from `authenticationOptions` to what you picked, `networkAuthenticationOptions`
3. update any test snapshots

Path two... the more thorough option
1. Delete all of `authenticationTypeOptions` between lines `32-51` and move the two valid options to be assigned to the new value you created like
   ```
   const networkAuthenticationOptions = [
      {
        title: () => t('form-dialog.label', { context: ['option', 'sshKey'] }),
        value: 'sshKey'
      },
      {
        title: () => t('form-dialog.label', { context: ['option', 'usernamePassword'] }),
        value: 'usernamePassword',
        selected: true
      }
   ]
   ```
3. Remove the `authenticationOptions` prop entirely
4. update any test snapshots

mirekdlugosz: There are two commits now, but they should be squashed before merge. I decided to split them ot make review easier, and to have something we can roll back to in case we decide we want to play it safer after all.

cdcabrera (approved):
lgtm!

PR #370, opened 2024-05-14
fix(viewToolbar): dsc-682 add missing types to filter

mirekdlugosz:
Credential type and Source type filter items were missing RHACS and Ansible. We must have forgot to add them back when we introduced these source types. Add these items now.

Relates to JIRA: DISCOVERY-682


cdcabrera (approved):
lgtm

bruno-fs (approved):

PR #377, opened 2024-06-10
test: ds-680 add unit tests to hooks/useAlerts

nicolearagao:
What's included
test: ds-680 add unit tests to hooks module
 
Notes
Covers useAlerts file

How to test
2. `$ npm test`

2. `$ npm run test:watch`

### Local run check
2. `$ npm run start`

### Check the build
2. `$ npm run build`

Updates issue/story

### DIFF
--- a/src/hooks/__tests__/useAlerts.test.ts
+++ b/src/hooks/__tests__/useAlerts.test.ts
@@ -1,0 +1,56 @@
+import { AlertProps } from '@patternfly/react-core';
+import { act, renderHook } from '@testing-library/react';
+import { useAlerts } from '../useAlerts';
+
+describe('useAlerts', () => {
+  it('should add an alert and verify it is in the alerts list', () => {
+    const { result } = renderHook(() => useAlerts());
+
+    act(() => {
+      result.current.addAlert({ title: 'Test Alert', variant: 'success', id: 'alert-1' });
+    });
+
+    expect(result.current.alerts).toMatchSnapshot('single alert list');
+  });
+
+  it('should add multiple alerts and verify all are present in the alerts list', () => {
+    const { result } = renderHook(() => useAlerts());
+    const alerts: AlertProps[] = [
+      { id: 'lorem', title: 'ipsum', variant: 'success' },
+      { id: 'hello', title: 'world', variant: 'danger' },
+      { id: 'dolor', title: 'sit', variant: 'success' }
+    ];
+
+    act(() => {
+      result.current.addAlert(alerts);
+    });
+
+    expect(result.current.alerts).toEqual(alerts);
+  });
+
+  it('should allow removing alerts by any property', () => {
+    const { result } = renderHook(() => useAlerts());
+
+    const alerts: AlertProps[] = [
+      { id: 'lorem', title: 'ipsum', variant: 'success' },
+      { id: 'hello', title: 'world', variant: 'danger' },
+      { id: 'dolor', title: 'sit', variant: 'success' },
+      { id: 'amet', title: 'consectetur', variant: 'danger' },
+      { id: 'adipiscing', title: 'elit', variant: 'danger' }
+    ];
+
+    act(() => {
+      result.current.addAlert(alerts);
+    });
+
+    act(() => {
+      result.current.removeAlert(alerts[0].title);
+      result.current.removeAlert(alerts[1].id);
+      result.current.removeAlert('nonExistentValue');
+      result.current.removeAlert('danger');
+    });
+
+    expect(result.current.alerts).toMatchSnapshot('removed alerts list');
+    expect(result.current.alerts).toHaveLength(1);
+  });
+});
--- a/src/hooks/useAlerts.ts
+++ b/src/hooks/useAlerts.ts
@@ -1,43 +1,39 @@
 /**
- * This custom React Hook provides functions to manage alerts in your application.
- * It allows you to add and remove alerts with specific titles, variants, and unique keys.
+ * This Hook provides functions to manage alerts.
+ * It allows you to add and remove alerts using PatternFly's AlertProps and unique IDs.
  * @module useAlerts
  */
 import React from 'react';
 import { AlertProps } from '@patternfly/react-core';
 
 const useAlerts = () => {
-  const [alert, setAlerts] = React.useState<Partial<AlertProps>[]>([]);
+  const [alerts, setAlerts] = React.useState<AlertProps[]>([]);
 
   /**
-   * Add an Alert
+   * Adds a new alert or multiple alerts to the list.
    *
-   * This function adds an alert to the list of alerts.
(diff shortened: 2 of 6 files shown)


cdcabrera (commented):
@nicolearagao something to consider... a new issue for a useAlerts/Alerts component that combines the useAlerts hook and the `AlertGroup` that is currently under each view. Really those 2 things should be grouped into a centralized component, it would be less error prone on implementation, and generally less work in the long run.

cdcabrera (changes requested):

src/hooks/useAlerts.ts
```diff
@@ -1,47 +1,43 @@
...
    *
-   * @param {AlertProps} options - An object containing the properties of the alert.
-   * @param {string} options.title - The title or content of the alert.
-   * @param {AlertProps['variant']} options.variant - The variant or style of the alert (e.g., 'success', 'danger').
-   * @param {React.Key} options.key - A unique key to identify the alert.
+   * @param {AlertProps} options - The properties of the alert.
    */
   const addAlert = (options: AlertProps) => {
-    setAlerts(prevAlerts => [...prevAlerts, options]);
+    const uniqueId = `alert-${alertCounter++}`;
```
nicolearagao, line 22: @cdcabrera I used an alertCounter here to generate the ID mainly for its simplicity, but also due to the risk of id collision of getUniqueId from PF. I mean, this use case is really simple and maybe I shouldn't worry about this level of complexity here. What's your opinion on the matter ?

cdcabrera: So this is where default properties would normally come in at the component level.
> Generally, I'm not a fan of sequential identifiers, history has shown me there are too many things that end up going weird.

We're attempting to bypass the need to write a centralized alert component by modifying state only. The preference is we do those defaults at the component level. What you have would sort-of work, but it leans towards modifying state. It's not wrong, we just need to be consistent since right now the code is blending state modifications and component defaults.
> This state modification behavior is why v1 code wraps many PF components, PF components don't always come with state, some do, but not all.

So, we need to pick a path, picking a path includes.... (I can go either path on this, so no wrong choice, but you have to make that decision)

**"fix what we can and get the tests working" aka continue what you started**
- we continuing your work to separate out the `useAlerts` hook (we probably will end up doing this for each hook in case they cover similar future modifications)
- we continue minimally modifying `useAlerts` which includes
   - continuing to move over to removing "key" and flipping to "id"
   - but we modify where the `getUniqueId` is placed, we move it to the component level instead, in your views
      ```
      <AlertGroup isToast isLiveRegion>
        {alerts.map(({ id, variant, title }) => (
          <Alert
            timeout={8000}
            onTimeout={() => id && removeAlert(id)}
            variant={AlertVariant[variant || 'info']}
            title={title}
            actionClose={
              <AlertActionCloseButton
                title={title as string}
                variantLabel={`${variant} alert`}
                onClose={() => id && removeAlert(id)}
              />
            }
            key={id || getUniqueId()}
          />
        ))}
      </AlertGroup>
      ```

**picking path "fix all alerts now" includes**

 - we create a centralized "alert component" similar to what we created in v1 and co-locate a new component possibly with the hook it needs to function
    - https://github.com/quipucords/quipucords-ui/blob/main/src/components/toastNotificationsList/toastNotificationsList.js
- we update all of our views with this new component and remove the unique activations of alerts per view
- we update our tests

cdcabrera (changes requested):

src/hooks/__tests__/useAlerts.test.ts
```diff
@@ -0,0 +1,83 @@
...
+    const alertCount = 3;
+
+    act(() => {
+      for (let i = 1; i <= alertCount; i++) {
+        result.current.addAlert({
+          id: `alert-${i}`,
+          title: `Test Alert ${i}`,
+          variant: i === 2 ? 'danger' : 'success'
+        });
+      }
```
cdcabrera, line 26: optional: (i highly recommend this)
you highlighted a use case here... the ability to apply multiple alerts at once. it might be a nice addition to simply allow the application of bulk messages without having to cycle through a listing, especially since it'd be an easier add here than in other areas of code
```
const addAlert = (options: AlertProps | AlertProps[]) => {
    setAlerts(prevAlerts => [...prevAlerts, ...((Array.isArray(options) && options) || [options])]);
  };
```

optional:
this for loop format isn't wrong, and it works, but it does come across as another typescript inflicted format... this can hide simplifications

```
  const { result } = renderHook(() => useAlerts());
    const alerts: AlertProps[] = [
      { id: 'lorem', title: 'ipsum', variant: 'success' },
      { id: 'hello', title: 'world', variant: 'danger' },
      { id: 'dolor', title: 'sit', variant: 'success' }
    ];

    act(() => {
      alerts.forEach(alert => result.current.addAlert(alert));
    });

  expect(result.current.alerts).toEqual(alerts);
```

and if you decided to also do the optional work to allow lists in addAlerts
```
  const { result } = renderHook(() => useAlerts());
    const alerts: AlertProps[] = [
      { id: 'lorem', title: 'ipsum', variant: 'success' },
      { id: 'hello', title: 'world', variant: 'danger' },
      { id: 'dolor', title: 'sit', variant: 'success' }
    ];

    act(() => {
      result.current.addAlert(alerts);
    });

    expect(result.current.alerts).toEqual(alerts);
```

src/hooks/__tests__/useAlerts.test.ts
```diff
@@ -0,0 +1,83 @@
...
+      result.current.addAlert({ title: 'Test Alert', variant: 'success' });
+    });
+
+    act(() => {
+      result.current.removeAlert('nonExistentValue');
+    });
+
+    expect(result.current.alerts).toHaveLength(1);
+  });
+});
```
cdcabrera, line 83: required: we can optimize the last 3 tests into a single `it`since technically we're doing the same thing just different properties. **Plus** we could include something additional, the ability to remove multiple alerts at once....

...using what you have, while we could use the returned "id" it feels more like a test if the list we put in really is the list we're getting out, let's make a mock instead...
```
  it('should allow removing alerts by any property', () => {
    const { result } = renderHook(() => useAlerts());

    // ...addAlertsHere

    act(() => {
      result.current.removeAlert(alerts[0].title); // leveraging the format like before
      result.current.removeAlert(alerts[1].id); // leveraging the format like before
      result.current.removeAlert('nonExistentValue'); // nonExistent like before
      result.current.removeAlert('danger'); // multiple alerts
    });

    // or you could do a snapshot, like expect(result.current.alerts).toMatchSnapshot('removed');
    expect(result.current.alerts).toHaveLength(1); 
   
  });
```

cdcabrera (changes requested):
test updates and useAlerts should be good to fully separate out

cdcabrera (approved):
lgtm!

PR #379, opened 2024-06-18
DS-680 hooks/useCredentialApi

nicolearagao:
What's included
test: ds-680 add unit tests to hooks module
 
Notes
useCredentialApi - Placeholder onDeleteSelectedCredentials was not tested, deletion in bulk hasn't been implemented(delete it ?)

How to test
2. `$ npm test`

2. `$ npm run test:watch`

### Local run check
2. `$ npm run start`

### Check the build
2. `$ npm run build`

Updates issue/story


cdcabrera (changes requested):
The unit tests for this hook highlight a few issues that should be resolved to help us allow for future expansion...

1. The hook function name does not accurately reflect what it accomplishes, deleting credentials, let's fix this
   - required change: Let's update the hook function name to some more specific, like `useDeleteCredentialsApi`
   - additional reason: we can have multiple hooks in one file, this makes it easier to test since a large function that returns multiple functions may have private internals that are awkward to test against. considering the file itself as the class, in some cases, may help with simplification 

1. The `useCredentialApi` hook function attempts to house state inside itself that directly affects the view component modal display with `setPendingCredential` instead of keeping it within the `viewCredentialsList.tsx` component. This is confusing. `setPendingCredential` state should be housed in relation to what component it is affecting to avoid confusion on why/where it's being used. Let's make it easier on ourselves...
   - required change: move the `pendingDeleteCred, setPendingDeleteCred` state to to `viewCredentialsList.tsx` and remove it from the hook. This will also involve removing the various references to `pendingDeleteCredential` inside the hook... see the next change for more detail
      ```
      const [pendingDeleteCredential, setPendingDeleteCredential] = React.useState<CredentialType>();
      ```
   - required change: let's pass credentials to the callbacks
      ```
         const onDeleteCredential = useCallback(
            (credential: CredentialType) => {
      ``` 
1. The hook function attempts to draw a distinction between deleting a single credential and multiple credentials, technically there is no difference from the API's perspective. Let's update the hook itself so it has the ability to delete a single and multiple credentials [similar to how v1 handles this same issue](https://github.com/quipucords/quipucords-ui/blob/main/src/redux/actions/credentialsActions.js#L29)

#### And finally
After those 3 updates... the end result of what we're going for should look something like the below. 

```
const useDeleteCredentialApi = () => {
  const { t } = useTranslation();
  const { addAlert } = useAlerts();

  // similar to what you've already started, consider a consistent name or prefix to help highlight this is only the base api call so we can apply it across other hooks
  const apiDeleteCredential = (id: CredentialType['id']) =>
    axios.delete(`${process.env.REACT_APP_CREDENTIALS_SERVICE}${id}/`);

  // delete a single, or multiple, credentials. And handle success, error, and cleanup operations.
  const onDeleteCredentials = useCallback(
    (credential: CredentialType | CredentialType[]) => {
      ...
      return Promise.all(updatedCredentials...
    [addAlert, apiDeleteCredential, t]
  );

  return {
    apiDeleteCredential,
    onDeleteCredentials
  };
};
```

This will also involve updating `viewCredentialsList.tsx` to be able to handle this updated format, where things like
   - the `then` and `catch` can be maintained in the above `onDeleteCredentials` hook callback
   - and `finally` could be used in the`viewCredentialsList.tsx` component for things like `refresh`
      - example: `onClick={() => onDeleteCredentials(selectedItems).finally(() => onRefresh())}`

src/hooks/__tests__/useCredentialApi.test.ts
```diff
@@ -0,0 +1,140 @@
...
+    };
+    (axios.post as jest.MockedFunction<typeof axios.post>).mockRejectedValue(error);
+
+    await act(async () => {
+      await result.current.deleteCredentials(mockCredential1);
+    });
+
+    expect(mockOnAddAlert).toHaveBeenCalledWith(expect.objectContaining({ variant: 'danger' }));
+  });
+});
```
cdcabrera, line 140: I compliment the thoroughness of these tests, however... we shouldn't be testing axios. Axios is already tested we should be testing what we're actually doing, submitting formatted data and displaying alerts

required: 
- I've opened up a PR that will provide you with some enhancements for testing, **bring that commit into this work**... we'll be merging that work before yours since it helps with your other PRs too
- let's focus your unit tests and try to make sure we house our mocks under the `describe`
   - we shouldn't actually need full fledge mock responses... 
- I've removed the detailed mocks and opted for simplistic data. The mocks look like they took time, apologies for removing them. The reason, we're not necessarily concerned about e2e level detail at this point is that these are unit tests and we are using Typescript. Typescript should help us catch issues at a detail level if our typings are setup correctly. We wrote functions that don't really care about anything but their piece, so from that perspective we just need enough to be to test that functionality and trust Typescript. 
- I've broken the tests out into something that uses a more simplistic axios mock using the `jest.spyOn` method where we'll detect the `calls` to with snapshots... we can apply this `calls` concept to the `onAddAlert` callback too

- And I've added direct tests for the success and error callbacks now that we're exporting them, and we're no longer bogged down trying to mock out the entire hook

```
describe('useDeleteCredentialApi', () => {
  let mockOnAddAlert;
  let hookResult;

  beforeEach(() => {
    mockOnAddAlert = jest.fn();
    const hook = renderHook(() => useDeleteCredentialApi(mockOnAddAlert));
    hookResult = hook?.result?.current;
  });

  afterEach(() => {
    jest.clearAllMocks();
  });

  // Let's keep the axios call simple, we can use a jest.spyOn instead, ignore the typings on spyOn they also appear out-of-date/busted similar to `toMatchSnapshot`
  // we're going to use `mock.calls` and pull out the "attempt" at an API response
  it('should attempt an api call to delete credentials', () => {
    const { apiCall } = hookResult;
    const mockAxios = jest.fn();
    jest.spyOn(axios, 'post').mockImplementation(mockAxios);

    apiCall([456, 789]);
    expect(mockAxios.mock.calls).toMatchSnapshot('apiCall');
  });

  it('should attempt to delete single, and multiple credentials', () => {
    const { deleteCredentials } = hookResult;
    const mockAxios = jest.fn();
    jest.spyOn(axios, 'post').mockImplementation(mockAxios);

    deleteCredentials({ id: 123 });
    deleteCredentials([{ id: 456 }, { id: 789 }]);
    expect(mockAxios.mock.calls).toMatchSnapshot('deleteCredentials');
  });

  it('should process an API success response', () => {
    const { callbackSuccess } = hookResult;

    // danger
    callbackSuccess(
      {
        data: {
          skipped: [{ credential: 123 }, { credential: 456 }]
        }
      },
      [
        { id: 123, name: 'Lorem' },
        { id: 456, name: 'Ipsum' }
      ]
    );

    // warning
    callbackSuccess(
      {
        data: {
          deleted: [456],
          skipped: [{ credential: 123 }]
        }
      },
      [
        { id: 123, name: 'Lorem' },
        { id: 456, name: 'Ipsum' }
      ]
    );

    // success
    callbackSuccess(
      {
        data: {
          deleted: [123, 456]
        }
      },
      [
        { id: 123, name: 'Lorem' },
        { id: 456, name: 'Ipsum' }
      ]
    );

    expect(mockOnAddAlert.mock.calls).toMatchSnapshot('callbackSuccess');
  });

  it('should process an API error response', () => {
    const { callbackError } = hookResult;

    callbackError(
      {
        response: {
          data: {
            detail: 'Lorem ipsum'
          }
        }
      },
      [
        { id: 123, name: 'Lorem' },
        { id: 456, name: 'Ipsum' }
      ]
    );

    callbackError(
      {
        response: {
          data: {
            message: 'Dolor sit'
          }
        }
      },
      [
        { id: 123, name: 'Lorem' },
        { id: 456, name: 'Ipsum' }
      ]
    );

    callbackError(
      {
        message: 'Amet'
      },
      [
        { id: 123, name: 'Lorem' },
        { id: 456, name: 'Ipsum' }
      ]
    );

    callbackError({}, [
      { id: 123, name: 'Lorem' },
      { id: 456, name: 'Ipsum' }
    ]);

    expect(mockOnAddAlert.mock.calls).toMatchSnapshot('callbackError');
  });
});
```

cdcabrera (changes requested):
looks good, you've helped expose optimizations which is a good thing!

In addition to the required changes _there are 2 extras that you should do._..

- Get rid of the Patternfly `getUniqueId` since it IS NOT test safe.. we originally had a `generateId` in the helper file similar to below, you could restore that, it is test safe. And it WILL help with testing snapshots if you replace all instances of it in `useCredentialApi.ts`
```
/**
 * Generate a random ID.
 *
 * @param {string} prefix
 * @returns {string}
 */
const generateId = (prefix = 'generatedid') =>
  `${prefix}-${(process.env.REACT_APP_ENV !== 'test' && Math.ceil(1e5 * Math.random())) || ''}`;
```

- Make addAlert have the ability to automatically generate an ID if you don't apply one. This will help us out in the long run so we don't have to continuously apply it every time we want to add an alert
```
const addAlert = (options: AlertProps | AlertProps[]) => {
    const updatedOptions = ((Array.isArray(options) && options) || [options]).map(options => ({
      ...options,
      id: options.id || helpers.generateId()
    }));

    setAlerts(prevAlerts => [...prevAlerts, ...updatedOptions]);
  };
```

src/hooks/useCredentialApi.ts
```diff
@@ -1,44 +1,170 @@
...
+        ?.map(({ credential }) => updatedCredentials.find(cred => credential === cred.id)?.name)
+        .filter(Boolean)
+        .join(', ');
+
+      const missingNames = data?.missing
+        ?.map(id => updatedCredentials.find(cred => id === cred.id)?.name)
+        .filter(Boolean)
+        .join(', ');
+
+      const missingCredsMsg = 'The following credentials could not be found:' + missingNames;
```
cdcabrera, line 61: required: let's be consistent and start avoiding old skool string concats, flip to js template/string literal
```
`The following ... : ${missingNames}`
```

src/hooks/useCredentialApi.ts
```diff
@@ -1,44 +1,170 @@
...
+
+      // All credentials were deleted successfully
+      if (data?.deleted?.length === updatedCredentials.length) {
+        const credentialCount = updatedCredentials.length;
+        const context = credentialCount === 1 ? 'deleted-credential' : 'deleted-credentials';
```
cdcabrera, line 66: required: we're working harder than we need to here, and in a few other places this looks like appears too.
i18next provides a mechanism for plurals already using `count`. let's flip here and in the other spots to `count`

> string display logic can get code verbose and confusing fast always try to shift it to i18next and the locale strings file where verbosity is expected

```
const successMessage = t('toast-notifications.description', {
   context: 'deleted-credential',
   count: updatedCredentials.length,
   name: updatedCredentials.map(({ name }) => name).join(', ')
});
```

in en.json starting on line `310` (you'll still need to confirm you don't have overlap I see another `other` already in the file on line `316`) 
```
change line 310: "description_deleted-credential": "Credential(s) were successfully deleted.",
add line: "description_deleted-credential_one": "Credential {{name}} was successfully deleted.",
change line 311: "description_deleted-credential_other": "Credentials {{name}} were successfully deleted."
...
remove and/or keep the copy if you like from line 316: "description_deleted-credential_other": "Deleted credentials {{name}} and more",
```

src/hooks/useCredentialApi.ts
```diff
@@ -1,44 +1,170 @@
...
+        return;
+      }
+
+      // No credentials deleted, either all skipped or all missing
+      if (
+        data?.skipped?.length === updatedCredentials.length ||
+        data?.missing?.length === updatedCredentials.length
+      ) {
+        if (data?.skipped?.length === updatedCredentials.length) {
+          const countValues = data?.skipped?.length === 1 ? 'Credential' : 'Credentials';
```
cdcabrera, line 85: required: move to use `count` in i18next. you can display `{{count}}` in your strings too!

src/hooks/useCredentialApi.ts
```diff
@@ -1,44 +1,170 @@
...
+        }
+        {
+          console.log(missingCredsMsg);
+          return;
+        }
+      }
+
+      // Some credentials deleted, some skipped or missing
+      if (data?.deleted?.length && data?.skipped?.length) {
+        const skippedCountValues = data?.skipped?.length === 1 ? 'Credential' : 'Credentials';
```
cdcabrera, line 106: required: move to use `count` in i18next. you can display `{{count}}` in your strings too!

cdcabrera (changes requested):
Looking better! We're getting close. String updates! 

Next round I'll post some logic updates since we may be working harder than we need to on our conditional checks, but i have to confirm

src/hooks/useCredentialApi.ts
```diff
@@ -1,44 +1,170 @@
...
+            credential_count: countValues
+          });
+          onAddAlert({
+            title: errorMessage,
+            variant: 'danger',
+            id: helpers.generateId()
+          });
+          return;
+        }
+        {
```
cdcabrera, line 98: required: this looks like a bug, extra or missing condition?

src/hooks/useCredentialApi.ts
```diff
@@ -1,44 +1,165 @@
...
+    },
+    [onAddAlert, t]
+  );
+
+  const callbackError = useCallback(
+    (
+      { message, response }: AxiosError<ApiDeleteCredentialErrorType>,
+      updatedCredentials: CredentialType[]
+    ) => {
+      const errorCountValues = updatedCredentials.length === 1 ? 'credential' : 'credentials';
```
cdcabrera, line 127: required: move to using i18next `count` instead

src/hooks/useCredentialApi.ts
```diff
@@ -1,44 +1,165 @@
...
+        onAddAlert({
+          title: successMessage,
+          variant: 'warning',
+          id: helpers.generateId()
+        });
+        if (data?.missing?.length) {
+          console.log(missingCredsMsg);
+          return;
+        }
+      }
```
cdcabrera, line 117: required: we have a unit test that is failing at displaying `success` messages on deletion, specifically, 
```
exports[`useDeleteCredentialApi should handle success while attempting to delete single and multiple credentials: deleteCredentials, success 1`] = []
```
This should be populated with messages...

**Snapshots are tricky**, they have be double checked manually the first few times to make sure the data is coming through. In this case your `success` messages aren't actually firing based on the logic you have

Recommendation
- Let's reset the logic, it appears we are working significantly harder than we need to. 
 
   Since our logic is concerned with things that aren't deleted let's refocus it on that. Let's try "process of elimination" and reverse the order so "success" is the last thing that happens. 
  
   What that means is we'll be limiting our conditions to prove "its not a success" since what we have currently is a bunch of messaging around that concept already, it's less conditions, less prone to developer error

```
      // No credentials deleted, either all skipped or all missing - Exact/Specific checks here are fine, let's do this first the OR gives us options. It's parallel to checking for "all deleted" but in the other direction
      if (
        data?.skipped?.length === updatedCredentials.length ||
        data?.missing?.length === updatedCredentials.length
      ) {
       ...
        onAddAlert({
          ...
        });

        if (data?.missing?.length) {
          console.log(missingCredsMsg);
        }
        return;
      }

      // Some credentials deleted, some skipped or missing - Truthy OR check instead, we don't really need to care beyond the existence of `skipped` or `missing` to fire the alerts. By their existence it means it didn't go according to plan
      if (data?.skipped?.length || data?.missing?.length) {
        ...
        onAddAlert({
          ...
        });

        if (data?.missing?.length) {
          console.log(missingCredsMsg);
        }
        return;
      }

     // All credentials were deleted successfully - **No conditional check necessary**. You can add one if you want, but it's unnecessary since we've eliminated all the skipped and missing before we got to this point
       ...
      onAddAlert({
        ...
      });
```

**OPTIONAL**
If you really wanted to take this even further... you started the concept of applying `missingCredsMsg` in the `console.log`... what's to stop you from doing that to the `onAddAlert` locale string, you could take that a step further and have a single `onAddAlert` at the end. the limitations could be seen as readability with the use of an `else if`... but it highlights that you can take even further... I see an additional optimization but I'll let you figure it out, the hint is ... you're doing string logic we could rely even more on the i18next translation/t function and just because "context" has a `property` doesn't mean it'll be used by the the translation/t function 

```
const message: AlertProps = {
        title: t('toast-notifications.description', {
          context: 'deleted-credential',
          count: updatedCredentials.length,
          name: updatedCredentials.map(({ name }) => name).join(', ')
        }),
        variant: 'success',
        id: helpers.generateId()
      };

// No credentials deleted, either all skipped or all missing
      if (
        data?.skipped?.length === updatedCredentials.length ||
        data?.missing?.length === updatedCredentials.length
      ) {
        message.title = t('toast-notifications.description', {
          context: 'skipped-credential',
          name: skippedNames,
          count: data?.skipped?.length
        });
        message.variant = 'danger';
      } else if (data?.skipped?.length || data?.missing?.length) {
        message.title = t('toast-notifications.description', {
          context: 'deleted-credentials-skipped-credentials',
          deleted_names: deletedNames,
          skipped_names: skippedNames,
          count: data?.skipped?.length
        });
        message.variant = 'warning';
      }

      // All credentials were deleted successfully
      onAddAlert(message);

      if (data?.missing?.length) {
          console.log(`The following credentials could not be found: ${missingNames}`);
      }

      return;
```

cdcabrera (changes requested):
Very close! Optimizations and a busted unit test

cdcabrera (approved):
lgtm!

PR #382, opened 2024-06-18
DS-680 hooks/useSearchParam

nicolearagao:
What's included
refactor: ds-680 remove unused hooks/useSearchParam

How to test
2. `$ npm test`

2. `$ npm run test:watch`

### Local run check
2. `$ npm run start`

### Check the build
2. `$ npm run build`

Updates issue/story


cdcabrera (commented):
The useSearchParam hook is an unused leftover. Unclear what a passive read of the browser level querystring was intended for in this application ui. 

The v1 code also doesn't have something like this so we should feel comfortable removing it for now and then if we have to we can pull it back. Keep your branch around

cdcabrera (approved):

PR #417, opened 2024-08-29
fix: fix broken apiStage.sh script

infinitewarp:
## What's included

Fix the `apiStage.sh` script as requested at standup on 2024-08-29.

Disclaimer: I do not use or maintain this script. In its current state in this branch, it nominally works on my machine (containers start, and I am able to access the running service in my local browser at http://127.0.0.1:8000), but that's as far as I will test or support this script.

    fix: add missing "proxy_set_header X-Forwarded-Proto $scheme"

    See: https://redhat-internal.slack.com/archives/C02QSNF1UKE/p1724871007554689

    fix: use correct env vars when starting quipucords server container

    See: https://github.com/quipucords/quipucords/pull/2726

    fix: start redis before starting quipucords server

    Modern versions of quipucords require redis.

    I do not use this script, and I do not support it.
    I consider this script to be a hack, and I absolve myself of any responsibility.
    Proper installations of quipucords should use quipucords-installer.
    Use this script at your own risk.

    fix: quipucords serves port 8000 not 443

    See: https://github.com/quipucords/quipucords/pull/2729

### DIFF
--- a/scripts/apiStage.sh
+++ b/scripts/apiStage.sh
@@ -3,8 +3,7 @@
 #
 # Quick check to see if a container is running
 #
-checkContainerRunning()
-{
+checkContainerRunning() {
   local CHECKONE=$1
   local COUNT=1
   local DURATION=10
@@ -14,7 +13,7 @@ checkContainerRunning()
 
   while [ $COUNT -le $DURATION ]; do
     sleep $DELAY
-    (( COUNT++ ))
+    ((COUNT++))
     if [ -z "$($PODMAN ps | grep $CHECKONE)" ]; then
       break
     fi
@@ -33,8 +32,7 @@ checkContainerRunning()
 #
 # Clean container tooling
 #
-cleanApi()
-{
+cleanApi() {
   echo "Cleaning everything, \$PODMAN and data..."
   printf "${RED}\n"
   $PODMAN system prune -f
@@ -44,16 +42,15 @@ cleanApi()
 #
 # Stop all QPC containers
 #
-stopApi()
-{
+stopApi() {
   $PODMAN stop -t 0 $($PODMAN ps --filter name="qpc*")
+  $PODMAN network rm -f -t 0 qpc-stage-network
 }
 #
 #
 # Run stage/dev setup, used for local development against the latest API, partially working login
 #
-stageApi()
-{
+stageApi() {
   local HOST=$1
   local PORT=$2
   local PASSWORD=$3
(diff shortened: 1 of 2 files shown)


scripts/apiStage.sh
```diff
@@ -64,6 +61,13 @@ stageApi()
   $PODMAN rm $NAME
   $PODMAN pull $CONTAINER
 
+  $PODMAN stop -t 0 qpc-redis
+  $PODMAN rm qpc-redis
+  $PODMAN pull registry.redhat.io/rhel9/redis-6:latest
```
cdcabrera, line 66: @infinitewarp Anyway there is a redis image that doesn't require logging in to the customer portal?

<insert comments about licenses and having to create or log into something to be able to use the work>

cdcabrera (changes requested):
I can not get this update to work, running the shell script directly produces a consistent error message regardless of whether I log into the customer portal

```
Error: initializing source docker://registry.redhat.io/rhel9/redis-6:latest: unable to retrieve auth token: invalid username/password: unauthorized: Please login to the Red Hat Registry using your Customer Portal credentials.
```

Based on the PR comments we should go over off to the side and work out what's supposed to happen and why this script exists

scripts/apiStage.sh
bruno-fs: if the image is available on the open registry, you can replace `registry.redhat.io` with `registry.access.redhat.com`.

if this doesn't work, then login is required.

infinitewarp: This is the image quipucords uses as defined here:
https://github.com/quipucords/quipucords-installer/blob/main/config/quipucords-redis.container
```
Image=registry.redhat.io/rhel9/redis-6:latest
```

Others certainly exist in docker hub, quay, and elsewhere, and you are welcome to change it at your own risk, but the dev team can't support others.

bruno-fs: I just confirmed and redis repo is not available in the open registry:

```
Error: initializing source docker://registry.access.redhat.com/rhel9/redis-6:latest: reading manifest latest in registry.access.redhat.com/rhel9/redis-6: unsupported: This repo requires terms acceptance and is only available on registry.redhat.io 
```

So, yeah, login is required.

cdcabrera: fixed by just moving to a docker hosted redis:6-alpine... good enough for development

cdcabrera (approved):
**this update keeps us moving in the short term given our internal dates for the first week of Sept on the UI** 

Minor mods moving over to the official redis:6-alpine image hosted on d<word>hub. I tried logging in many times to gain access to the RH image... no luck. Moving over to using Redis comes with caveats 

 

- we probably will end up having to move the image
- the **load time is slow**, like really really slow as everything spools up... you will get 401s 3s and possibly a variety of other errors, just be patient and let it load... then develop away

**...development should be open to all given our license on this work...**

bruno-fs:
>  ...development should be open to all given our license on this work...

Sure. Everyone can create JIRA tickets to add to commit messages /s

Different from JIRA, anyone can create a free account to get access to the images hosted on redhat registry.

PR #439, opened 2024-09-10
feat(credential): ds-439 add edit functionality

nicolearagao:
## What's included

feat(credential): ds-439 add edit functionality

### Notes
- edit addCredentialModal to render appropriate authType if existing credential
- add edit credential functionality to viewCredentialListView

1. `$ npm test`

### Local run check
1. `$ npm run start`

## Example

...
Screenshot from 2024-09-10 16-23-32
Screenshot from 2024-09-10 16-25-08
Screenshot from 2024-09-10 16-23-03
Screenshot from 2024-09-10 16-23-56

## Updates issue/story

[DISCOVERY-439](https://issues.redhat.com/browse/DISCOVERY-439)

### DIFF
--- a/src/views/credentials/__tests__/addCredentialModal.test.tsx
+++ b/src/views/credentials/__tests__/addCredentialModal.test.tsx
@@ -2,6 +2,7 @@ import React, { act } from 'react';
 import { render, renderHook, screen } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { shallowComponent } from '../../../../config/jest.setupTests';
+import helpers from '../../../helpers';
 import { AddCredentialModal, CredentialFormFields, useCredentialForm } from '../addCredentialModal';
 
 describe('AddCredentialModal', () => {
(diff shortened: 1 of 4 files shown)


src/views/credentials/__tests__/addCredentialModal.test.tsx
```diff
@@ -68,6 +69,29 @@ describe('useCredentialForm', () => {
     }).toMatchSnapshot('token auth');
   });
 
+  it('should derive auth type based on the provided credential', () => {
+    const mockCredential = {
+      id: 1,
+      name: 'Test Credential',
```
nicolearagao, line 75: @cdcabrera I know you don't like to use "complete" mocks in tests, and this mock specifically is problematic because every time we run it again a snapshot will fail because of the created_at, updated_at fields that are dynamic. I thought about adding a hard-coded date, but is there a better way to solve this problem ?

cdcabrera (changes requested):
Very close! Just a type fix

PR #442, opened 2024-09-12
fix: ds-356 axios response interceptors

cdcabrera:
## What's included

- fix: ds-356 axios response interceptors

### Notes
- assumed patch for token deletion, or random, 401s. this is an incremental change
- alerting users to 4xx level errors for login, similar to v1.8 is outside of this patch and possibly requires altering the current view component layout
- patching 4xx level errors with a login display message is outside of this patch
- patching 5xx level errors with a display message possibly requires alteration of current view base `get`s and is ignored with this patch

1. `$ npm test`
1. confirm linting and tests come back as expected

### Local run check
1. `$ npm run start:stage` or your equivalent use of tooling for previous determination of errors
1. attempt to generate a 401 for any one API call
   - expected result is bringing up the login screen on interaction with the UI
1. attempt to let the login token expire
   - expected result is bringing up the login screen on interaction with the UI

## Example

...

## Updates issue/story

[DISCOVERY-356](https://issues.redhat.com/browse/DISCOVERY-356)
ongoing

### DIFF
--- a/src/hooks/useLoginApi.ts
+++ b/src/hooks/useLoginApi.ts
@@ -151,7 +151,8 @@ const useUserApi = () => {
 /**
  * Get initial token. Apply and set token for all Axios request interceptors, global authorization
  */
-const useGetSetAuthApi = () => {
+const useGetSetAuthApi = ({ useLogout = useLogoutApi }: { useLogout?: typeof useLogoutApi } = {}) => {
+  const { callbackSuccess: revokeToken } = useLogout();
   const [isAuthorized, setIsAuthorized] = useState<boolean>(false);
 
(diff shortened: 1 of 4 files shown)


bruno-fs:
Nice. manually verified that a user that's already logged out (I did it in another browser) is properly logged out now.

src/hooks/useLoginApi.ts
```diff
@@ -151,7 +151,8 @@ const useUserApi = () => {
 /**
```
nicolearagao, line 151: future: move api calls to hook folder

nicolearagao (approved):
lgtm, works as intended.

PR #448, opened 2024-09-13
build: ds-775 integration checks moved to ci run only

cdcabrera:
## What's included

- build: ds-775 integration checks moved to ci run only

### Notes
- removes the developer level integration check for `build` and `build:brand` and limits it on the `test:ci-build`. 
   - existing build scripts at the image/container level should be able to use the original script without needing to update

### Coverage and basic unit test check
1. update the NPM packages with `$ npm install`
1. `$ npm test`
1. confirm the developer integration checks still fire with CI

### Check the build
1. update the NPM packages with `$ npm install`
1. `$ npm run build`
   - confirm developer integration checks do not run
1. `$ npm run build:brand`
   - confirm developer integration checks do not run
## Example

...

## Updates issue/story

[DISCOVERY-775](https://issues.redhat.com/browse/DISCOVERY-775)

PR #449, opened 2024-09-13
build: ds-776 move required packages to prod deps

cdcabrera:
## What's included

- build: ds-776 move required packages to prod deps

### Notes
**- this work requires #448 to function correctly**
- using both webpack and typescript has caveats, specifically around `tsloader`
   - `tsloader` attempts to load test files
   - `tsloader`ignores the `exclude` webpack configuration option. this creates a scenario where you either need a specific `tsconfig.json` for webpack or you need to dynamically update `tsconfig.json`. 
      - we attempted to run a parallel `tsconfig.webpack.json` configuration with negative results
      - we settled on using dynamic `tsconfig.json` update that applies the test file `exclude`, this consistently provides positive results 
      - because this "dynamic" update affects the developer level build checks we applied a `git restore` to the `tsconfig.json`. This is very similar to our previous use of rolling back Django templates
- pros/cons of doing this type of container build
   - pros 
      - limit packages that are "audited"
      - speeds up the build in general
   - cons
      - "unique scripting"

### Check that the original build remains the same
1. `$ npm run build`
   - confirm no errors happen
1. `$ npm run build:brand`
   - confirm no errors happen

### Check the prod packages only install of the build functions
1. delete your `node_modules` directory
1. run the prod packages only NPM install with `$ npm install --omit=dev`
1. `$ npm run build`
   - confirm no errors happen
1. `$ npm run build:brand`
   - confirm no errors happen

## Example

...

## Updates issue/story

[DISCOVERY-776](https://issues.redhat.com/browse/DISCOVERY-776)


bruno-fs:
Thanks for the hard work.

Lemme try build this downstream before we proceed with merge - hopefully I shall have a feedback tomorrow.

bruno-fs (approved):
Validated downstream with brew. LGTM

PR #454, opened 2024-09-18
style(cred,source,scan): ds-438  popup style fix

nicolearagao:
Unify popup style across Sources, Creds, Scans
Remove cancel button

### Notes
- fix to "[Mirek] On “Sources” list. I can click a number in “Credentials” column and that will open new popup window. That window has both “close” button and “cancel” link (all lowercase). “close” is with red background, as if that was destructive action. That popup should probably use the same style as “Last connected” popup - it is purely informative.
The same applies to popup that appears on “Credentials” page, when clicking number in “Sources” column
It also appears on “Scans” page, when clicking number in “Sources” column"

1. `$ npm test`

### Local run check
1. `$ npm run start`

## Example

Screenshot from 2024-09-18 13-44-29
Screenshot from 2024-09-18 13-42-12
Screenshot from 2024-09-18 13-42-03

## Updates issue/story

### DIFF
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -156,6 +156,7 @@
     "label_auth_cell_authToken": "Token",
     "label_auth_cell_sshKey": "SSH Key",
     "label_auth_tooltip": "Authorization type",
+    "label_close": "Close",
     "label_delete": "Delete",
     "label_edit": "Edit",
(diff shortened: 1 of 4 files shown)


cdcabrera (approved):
technical implementation lgtm!

PR #455, opened 2024-09-18
fix(credential):ds-438 use auth_type sent from api

nicolearagao:
Utilize newly added auth_type field sent by the API instead of inferring it in the front end 

### Notes
- [Related server side PR ](https://github.com/quipucords/quipucords/pull/2735)
- fix to "[Brad] If you start by running the old UI, create a network credential in the old UI using an SSH key (file path), and then start running the new UI, the credentials page crashes and turns completely blank.
2024-09-18 In progress PR
Old behavior: You can see all credentials.
Expected behavior: You can see all credentials that were created with the old UI in the new UI.
Current behavior: The credentials page crashes and shows a completely blank page (no UI or HTML/DOM at all).
[Slack thread with details](https://redhat-internal.slack.com/archives/C02QSNF1UKE/p1726001863082449?thread_ts=1726001806.711239&cid=C02QSNF1UKE)

1. `$ npm test`

### Local run check
1. `$ npm run start`

## Example

Screenshot from 2024-09-18 13-27-17

## Updates issue/story

### DIFF
--- a/src/helpers/__tests__/__snapshots__/helpers.test.ts.snap
+++ b/src/helpers/__tests__/__snapshots__/helpers.test.ts.snap
@@ -5,6 +5,9 @@ exports[`getAuthType should return a credential type: credentialTypes 1`] = `
   "Username and Password",
   "Token",
   "SSH Key",
+  "SSH Key File",
+  "Unknown credential type",
+  "Unknown credential type",
 ]
 `;
 
--- a/src/helpers/__tests__/helpers.test.ts
+++ b/src/helpers/__tests__/helpers.test.ts
@@ -48,21 +48,19 @@ describe('getAuthType', () => {
       become_user: undefined,
       become_password: undefined,
       sources: [],
+      auth_type: undefined,
       ...partialCredential
     });
 
   it(`should return a credential type`, () => {
     expect([
-      generateAuthType({ username: 'testUser', password: 'testPassword' }),
-      generateAuthType({ auth_token: 'mockToken' }),
-      generateAuthType({ ssh_key: 'mockSSH' })
+      generateAuthType({ auth_type: 'password' }),
+      generateAuthType({ auth_type: 'auth_token' }),
+      generateAuthType({ auth_type: 'ssh_key' }),
+      generateAuthType({ auth_type: 'ssh_keyfile' }),
+      generateAuthType({ auth_type: 'lorem' }),
+      generateAuthType({ auth_type: 'ipsum' })
     ]).toMatchSnapshot('credentialTypes');
-  });
-
-  it('should throw an error when credential has no authentication information', () => {
-    expect(() => {
-      generateAuthType({});
-    }).toThrow('Unknown credential type');
   });
 });
 
--- a/src/helpers/helpers.ts
+++ b/src/helpers/helpers.ts
@@ -87,7 +87,9 @@ const normalizeTotal = (
 enum authType {
   UsernameAndPassword = 'Username and Password',
   Token = 'Token',
-  SSHKey = 'SSH Key'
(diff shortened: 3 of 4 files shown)


src/helpers/helpers.ts
```diff
@@ -96,17 +97,19 @@ enum authType {
...
+    case 'password':
+      return authType.UsernameAndPassword;
+    case 'auth_token':
+      return authType.Token;
+    case 'ssh_key':
+      return authType.SSHKey;
+    case 'ssh_keyfile':
+      return authType.SSHKeyFile;
+    default:
+      throw new Error('Unknown credential type');
```
bruno-fs, line 111: Will this error be handled anywhere? If not, change the behavior here.

This is usually an OK behavior for server side, but not for UI - especially when you are NOT dealing with user input.

bruno-fs (changes requested):

src/helpers/__tests__/helpers.test.ts
```diff
@@ -48,14 +48,16 @@ describe('getAuthType', () => {
...
 
   it(`should return a credential type`, () => {
     expect([
-      generateAuthType({ username: 'testUser', password: 'testPassword' }),
-      generateAuthType({ auth_token: 'mockToken' }),
-      generateAuthType({ ssh_key: 'mockSSH' })
+      generateAuthType({ auth_type: 'password' }),
+      generateAuthType({ auth_type: 'auth_token' }),
+      generateAuthType({ auth_type: 'ssh_key' }),
+      generateAuthType({ auth_type: 'ssh_keyfile' })
```
cdcabrera, line 60: optional:
since we've refined everything down to `auth_type` you've opened the ability to loop through the enum values instead of declaring them individually... 

pseudo code
```
Object.values(helpers.authType).map(authType => generateAuthType({ auth_type: authType}))
```

cdcabrera (approved):
1:1 for ~solving the issue~... technical implementation lgtm!

src/helpers/helpers.ts
cdcabrera: this was existing... if it wasn't called out before I vote for iterating on it

cdcabrera: looking back through the slack thread Brad posted the error, yes the throw is the issue...

cdcabrera (changes requested):
per @bruno-fs the throw needs to be replaced with whatever fallback the team wants

src/helpers/helpers.ts
bruno-fs: I'm happy with the new code. I'll leave unresolved just in case @cdcabrera wants to add anything else, otherwise feel free to solve this thread.

bruno-fs (approved):
lgtm (but I didn't test the code, so take my review with a grain of salt)

infinitewarp (approved):

src/helpers/__tests__/__snapshots__/helpers.test.ts.snap
```diff
@@ -5,6 +5,9 @@ exports[`getAuthType should return a credential type: credentialTypes 1`] = `
   "Username and Password",
   "Token",
   "SSH Key",
+  "SSH Key File",
+  "Unknown credential type",
+  "Unknown credential type",
```
ruda, line 10: Is this duplicated "Unknow cred..." entry necessary?

bruno-fs: on the test we see 

```js
      generateAuthType({ auth_type: 'lorem' }),
      generateAuthType({ auth_type: 'ipsum' })
```

so yeah, that's the expected response

cdcabrera (approved):
throw appears resolved!

PR #457, opened 2024-09-19
fix(creds,sources): ds-438 names inconsistency

nicolearagao:
Follow consistent pattern in displaying credentials and sources names. 

### Notes
- fix "[Mirek] On Credentials page, when filtering by Credential Type, items in dropdown use different labels than “Type” column in a table:
Filter says “vCenter”, but table says “vCenter Server”
Filter says “Ansible”, but table says “Ansible Controller”
Other labels are consistent
The same applies to Sources page and Source Type
[Mirek] On Credentials page, when filtering by Credential Type, items in dropdown are sorted alphabetically. But when clicking “Add Credentials”, the same set of items is sorted in arbitrary order (Network on top, Ansible last). They probably should be consistent
The same applies to Sources page and Source Type filter"

1. `$ npm test`

### Local run check
1. `$ npm run start`

## Example

Screenshot from 2024-09-19 11-44-28
Screenshot from 2024-09-19 11-44-07
Screenshot from 2024-09-20 20-36-38
Screenshot from 2024-09-20 20-36-31

## Updates issue/story

### DIFF
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -317,7 +317,7 @@
   "dataSource": {
     "rhacs": "RHACS",
     "ansible": "Ansible Controller",
-    "network": "Network",
+    "network": "Network range",
     "openshift": "OpenShift",
     "satellite": "Satellite",
     "vcenter": "vCenter Server"
--- a/src/views/credentials/viewCredentialsList.tsx
+++ b/src/views/credentials/viewCredentialsList.tsx
@@ -118,32 +118,32 @@ const CredentialsListView: React.FunctionComponent = () => {
             {
               key: API_DATA_SOURCE_TYPES.ANSIBLE,
               label: API_DATA_SOURCE_TYPES.ANSIBLE,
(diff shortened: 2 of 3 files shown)


cdcabrera (approved):
appears consistent for the immediate issue. if the old string keys are no longer used may consider cleaning them up

A possible future or alternative approach.... 
Relying on static order works, but is obviously prone to the next developer changing it. 

Couple of alternatives
- since the translate function is dumping strings into your list you could have also applied a simple JS `.sort()` on the array/list
- you could also apply a `.sort()` on the values inside `simpleDropdown`

nicolearagao (commented):
As discussion with Yichen:
- filter in alphabetical order 
- add (creds/source) displayed as in sketch

ps: print screen for filter is not updated. The same translation key is used for both so it's network range not network. 
@mirekdlugosz  @cdcabrera 
Screenshot from 2024-09-20 10-24-15

PR #458, opened 2024-09-19
fix(scans): ds-437 remove redudant toast from scan

nicolearagao:
Remove redundant toast message generated by opening Last Scanned modal from scans view.

### Notes
- fix "[Mirek] On Scans page, clicking a link in “Last Scanned” column displays new modal (good) AND a toast notification with “Connection information displayed successfully.”. That toast seems to be development leftover which should be removed"

1. `$ npm test`

### Local run check
1. `$ npm run start`

## Example

[Screencast from 2024-09-19 14-54-56.webm](https://github.com/user-attachments/assets/b2fd818f-a74a-4b17-acc9-e639061efa48)

## Updates issue/story


src/hooks/useScanApi.ts
```diff
@@ -242,19 +242,9 @@ const useGetScanJobsApi = (onAddAlert: (alert: AlertProps) => void) => {
     []
   );
 
-  const callbackSuccess = useCallback(
-    (response: AxiosResponse<ScanJobsResponse>) => {
-      onAddAlert({
-        title: t('toast-notifications.description', {
-          context: 'scan-jobs_fetched',
```
mirekdlugosz, line 249: Should that string be also removed from `public/locales/en.json`? As far as I can tell, this is the only place it is used (but I might be wrong!).

mirekdlugosz (approved):
Looks good, added a question / suggestion. Not really blocking.

src/hooks/useScanApi.ts
nicolearagao: Yes, agreed! I added the removal of that string to the commit.

src/hooks/__tests__/useScanApi.test.ts
```diff
@@ -295,18 +295,6 @@ describe('useGetScanJobsApi', () => {
...
-
-    callbackSuccess({
-      data: {
-        results: [{ name: 'Lorem', id: '1' }]
-      }
-    });
-
-    expect(mockOnAddAlert.mock.calls).toMatchSnapshot('callbackSuccess');
-  });
-
```
cdcabrera, line 309: required: this test should be updated instead of removed to maintain consistency in our API hooks... the alternative approaches for when there are no toasts are in a few places... for example

- the `useShowConnectionsApi` test located around line `50` of this file... 
- the `useUserApi` test located in the `useLoginApi.test.ts` around line `160`...

cdcabrera (changes requested):
consistency in unit tests .. update the test instead of removing it

src/hooks/__tests__/useScanApi.test.ts
nicolearagao: pushed fixup!

cdcabrera: added a minor clean up commit

cdcabrera (approved):

PR #460, opened 2024-09-21
refactor(addSourceModal): align form state with credential form

nicolearagao:
Align form state with credential form in order to maintain consistency, but also to solve some of the state problems we are currently experiencing in the modal. 

### Notes
- fix to [Mirek] New source popup remembers credentials and “Verify SSL Certificate” state across closing and opening
easy way to reproduce: start creating a new source of any type. Select some credentials. Close the popup (Esc key or Cancel link). Now start creating a new source of different type. Credentials field will say that you have some credentials selected, but since they are of different type, they won’t appear on the list. If you try to submit form now, server will return error
- [Mirek] New OpenShift source modal. Credentials dropdown is multi-select, but if I try to submit OpenShift with many credentials, server will reject it saying that this source type can have only one credential.
Not tested, but I assume the same applies to virtually all source types except for Network
The only type of source that accepts multiple credentials is Network.
-  [Mirek] When editing a source, following fields have their default values, ignoring what was really in the source when it was created: 
Network: “Connect using Paramiko instead of Open SSH” (reverts to unchecked)
OpenShift: “Connection” (reverts to SSLv23) ; “Verify SSL Certificate” (reverts to checked) 
RHACS: “Connection” (reverts to SSLv23) ; “Verify SSL Certificate” (reverts to checked)
Satellite: “Connection” (reverts to SSLv23) ; “Verify SSL Certificate” (reverts to checked)
VCenter: “Connection” (reverts to SSLv23) ; “Verify SSL Certificate” (reverts to checked)
Ansible: “Connection” (reverts to SSLv23) ; “Verify SSL Certificate” (reverts to checked)
- [Mirek] When editing a source (any type), modal title says “Edit source: undefined”

1. `$ npm test`

### Local run check
1. `$ npm run start`

## Example

[Screencast from 2024-10-07 13-11-35.webm](https://github.com/user-attachments/assets/8dd60810-5b82-4c12-9900-89b2e56ec086)
[Screencast from 2024-10-07 13-12-56.webm](https://github.com/user-attachments/assets/f837ef98-61a6-4dc0-8a30-421cc60ff953)
[Screencast from 2024-10-07 13-13-44.webm](https://github.com/user-attachments/assets/18f70142-07eb-44c6-8c5d-6efcbaf6129f)
[Screencast from 2024-10-07 13-14-21.webm](https://github.com/user-attachments/assets/aed7ccc1-f27e-4ade-a5d0-87649f738e14)

## Updates issue/story

[DISCOVERY-436](https://issues.redhat.com/browse/DISCOVERY-436)

### DIFF
--- a/src/views/sources/addSourceModal.tsx
+++ b/src/views/sources/addSourceModal.tsx
@@ -14,6 +14,7 @@ import {
   Form,
   FormGroup,
   HelperText,
+  HelperTextItem,
   Modal,
   ModalVariant,
   TextArea,
@@ -174,20 +175,19 @@ const SourceForm: React.FC<SourceFormProps> = ({
       <FormGroup label="Credentials" fieldId="credentials" isRequired>
         <TypeaheadCheckboxes
           onChange={(selections: string[]) => {
-            const selectedIds = selections.map(Number);
-            const validIds = selectedIds.filter(id => !isNaN(id));
-
-            // If the user selects more than one credential, display a warning
-            if (!isNetwork && validIds.length > 1) {
-              alert('Only one credential can be selected for this source type.');
-            } else {
-              handleInputChange('credentials', validIds);
-            }
+            const validIds = selections.map(Number).filter(id => !isNaN(id));
+            handleInputChange('credentials', validIds);
           }}
           options={credOptions}
           selectedOptions={formData?.credentials?.map(String) || []}
           menuToggleOuiaId="add_credentials_select"
+          maxSelections={isNetwork ? Infinity : 1} // Limit selection to 1 for non-network sources
         />
+        {!isNetwork && (
+          <HelperText>
+            <HelperTextItem variant="warning">Only one credential can be selected for this source type.</HelperTextItem>
+          </HelperText>
+        )}
       </FormGroup>
       {isNetwork ? (
         <React.Fragment>
--- a/src/views/sources/addSourceModal.tsx
+++ b/src/views/sources/addSourceModal.tsx
@@ -6,13 +6,12 @@
  *
  * @module addSourceModal
  */
-import React, { useEffect, useState } from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import {
   ActionGroup,
   Button,
   Checkbox,
   Form,
-  FormContextProvider,
   FormGroup,
   HelperText,
   Modal,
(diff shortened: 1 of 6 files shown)


src/views/sources/addSourceModal.tsx
```diff
@@ -70,188 +105,220 @@ const AddSourceModal: React.FC<AddSourceModalProps> = ({
...
+  sourceType,
+  sourceName,
+  onClose = () => {},
+  onSubmit = () => {}
+}) => {
   return (
     <Modal
       variant={ModalVariant.small}
-      title={`${source ? 'Edit' : 'Add'} source: ${sourceType}`}
+      title={source ? `Edit Source: ${sourceName || ''}` : `Add Source: ${sourceType || ''}`}
```
cdcabrera, line 315: optional:
Declaring the name separate is definitely one path, it works! But we may be working harder than needed.

Since all we're doing when we call this component is using `sourceBeingEdited?.name` anyway it feels awkward to do that breakout before instead of doing it inside the component(s), like we do for editing. Doubly so since TS should be calling out any required props with our `SourceType`. 

If the concern is the optional chaining for `sourceBeingEdited?.name`, or the unit testing, I think we're still good since that application is more around the whole source being `undefined` rather than the attribute/prop being optional. And for the existing unit test it looks like that's already taken into account as a required prop in the mock.

```
title={source ? `Edit Source: ${source.name}` : `Add Source: ${sourceType || ''}`}
```

and if you don't like the old-skool format...

```
title={(source && `Edit Source: ${source.name}`) || `Add Source: ${sourceType || ''}`}
```

and if we want to adhere to graceful fallbacks for all responses =)

```
title={(source && `Edit Source: ${source.name || ''}`) || `Add Source: ${sourceType || ''}`}
```

cdcabrera (changes requested):
looks good! Minor changes, one optional

ruda:
I verified your changes, the results:

> [Mirek] New source popup remembers credentials and “Verify SSL Certificate” state across closing and opening...

Fixed: I'm no longer seeing it repeating.

> [Mirek] New OpenShift source modal. Credentials dropdown is multi-select, but if I try to submit OpenShift with many credentials, server will reject ...

Tested with OpenShift and Satellite sources. Not OK! OR FLAKY to me :(

When I started my test...

When you add or edit a new source, if you select one credential and then select another credential, the interface will tell you: `Only one credential can be selected for this source type.` After you click OK, you can still see the two credentials in a checkbox. After saving (both kept selected), just one credential is store.

But after a time, I'm no longer seeing this message. The behavior now is the message is not been showing. It looks like the 2 creds are added, but in the end, just one is accepted.

> [Mirek] When editing a source, following fields have their default values, ignoring what was really in the source... (connect/connection/sslcert/)

Fixed: it now respects the choices that has been made.

> [Mirek] When editing a source (any type), modal title says “Edit source: undefined”

Fixed: it now displays the source name. Ex. `Edit Source: anssrc`.

ruda:
There is a BUG.

* Select `Sources`. I have a few sources configured. All sources displays the number of credentials they are using.
* Edit any source.
* The Edit source modal appears.
* There is no credentials selected / zero credentials appearing.

What I expected: the interface should display the credential(s) that belongs to the source at the moment I'm editing it.

nicolearagao:
> There is a BUG.
> 
> * Select `Sources`. I have a few sources configured. All sources displays the number of credentials they are using.
> * Edit any source.
> * The Edit source modal appears.
> * There is no credentials selected / zero credentials appearing.
> 
> What I expected: the interface should display the credential(s) that belongs to the source at the moment I'm editing it.

@ruda This issue specifically is being fixed on a different PR. It was one of the items in the list and I am currently working on it. =)

cdcabrera (approved):
code work looks good!

nicolearagao (commented):
@ruda I changed the approach on how to handle this use case. I don't display the alert anymore. I added a helper text to all non-network sources, and I prevent the user to click on multiple credentials as well. 
Screenshot from 2024-10-10 15-35-21
Screenshot from 2024-10-10 15-35-11

@cdcabrera I tried to stick to the previous implemention to keep it simple, but I couldn't find a way to prevent the user to click on the second credential without over complicating it. I thought it wouldn't hurt to modify the component itself to support one more prop, as I definetely see us using this again in the future for other implementations.

src/views/sources/addSourceModal.tsx
```diff
@@ -31,34 +31,69 @@ interface AddSourceModalProps {
...
+  const typeValue = source?.source_type || sourceType?.split(' ')?.shift()?.toLowerCase();
+  const isNetwork = typeValue === 'network';
 
-  const sourceTypeValue = source?.source_type || sourceType?.split(' ')?.shift()?.toLowerCase();
-  const isNetwork = sourceTypeValue === 'network';
+  // Edit props, reset state on unmount
+  useEffect(() => {
+    if (source) {
+      setFormData({
+        credentials: source?.credentials?.map(c => c.id) || [],
```
mirekdlugosz, line 77: Do we need that `source?` everywhere? This entire block is inside `if (source)`, so I would assume we can count on `source` being defined.

src/views/sources/addSourceModal.tsx
```diff
@@ -70,188 +105,216 @@ const AddSourceModal: React.FC<AddSourceModalProps> = ({
...
+  );
+
+  const filterFormData = useCallback(
+    (data = formData) => {
+      const { credentials, useParamiko, sslVerify, sslProtocol, name, hosts, port } = data;
+      return {
+        name: name,
+        credentials: credentials?.map(c => Number(c)),
+        hosts: hosts?.split(','),
+        port: port || (isNetwork ? '22' : '443'),
```
mirekdlugosz, line 124: Sorry for bringing it so late and in this venue, but the default ports are:

* [`22` for Network scan](https://github.com/quipucords/quipucords/blob/main/quipucords/api/source/serializer.py#L525)
* [`6443` for OpenShift](https://github.com/quipucords/quipucords/blob/main/quipucords/api/source/serializer.py#L537)
* [`443` for everything else](https://github.com/quipucords/quipucords/blob/main/quipucords/api/source/serializer.py#L539)

I think this logic needs to be a little more complex. And probably in some other places where we deal with ports, too.

tests/__snapshots__/dist.test.ts.snap
```diff
@@ -28,9 +28,9 @@ exports[`Build distribution should match a specific file output: dist-output 1`]
   "./build/fonts/e0f30d27bec7c8dbb6fc.woff2",
   "./build/fonts/e4faf134e53e0f37296b.woff2",
   "./build/fonts/ed292ec87cce0d76fd3c.woff2",
-  "./build/images/032b764e59d78db07b56.svg",
+  "./build/images/133911d429584887f5c7.svg",
+  "./build/images/4ca682d556b7bf59f5c6.svg",
```
mirekdlugosz, line 32: That change was also made in #474 , which is now merged. Please pull and rebase.

src/views/sources/addSourceModal.tsx
nicolearagao: we actually do, we use this for both adding and editing sources. we set the fields we would like to be returned, but if we are creating a new source we will set the fields with empy values : setFormData(initialFormState);

src/views/sources/addSourceModal.tsx
nicolearagao: @ mirek If you are ok with this, I will add your observation to the docs and I will fix it in another PR. Because of the major changes we are doing here, it is "blocking" work in other PRs. I would like to get it merged as soon as possible.

mirekdlugosz: Sure, documenting and fixing later is fine! I think we all forgot that OpenShift uses special default port.

nicolearagao: added to the doc, will resolve conversation, ok?

tests/__snapshots__/dist.test.ts.snap
nicolearagao: done

mirekdlugosz (approved):

PR #469, opened 2024-10-04
fix: ds-356 login loading, error messages

cdcabrera:
## What's included

- fix: ds-356 login loading, error messages

### Notes
- error messaging for login was previously more integrated with the Django templates. moving to the token service does not provide the same level of error messages
- went ahead and provide the immediate response for 400, bad request, level status since that's what's provided by the API error response
   - this starts the process of applying error codes, via status codes, to UI locale strings

1. `$ npm test`
1. confirm tests now come back clean

### Local run check
1. `$ npm run start:stage`
1. confirm login error messages attempt to display on the login

## Example

Screenshot 2024-10-11 at 2 04 56 PM

## Updates issue/story

 [DISCOVERY-356](https://issues.redhat.com/browse/DISCOVERY-356)

### DIFF
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -124,6 +124,7 @@
   },
   "login": {
     "description": "Welcome! This inspection and reporting tool is designed to identify and report environment data, or facts, such as the number of physical and virtual systems on a network, their operating systems, and other configuration data.",
+    "error_400": "Unable to log in with provided credentials.",
     "label_login": "Log in",
     "label_username": "Username",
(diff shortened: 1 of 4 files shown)


nicolearagao (approved):
This looks good so far. We gotta do work on backend to bring the correct errors forward, so  I think we can iterate on this when that is done.

cdcabrera:
Agreed @nicolearagao ! I'll rebase and provide minimal directions for testing for tracking

PR #475, opened 2024-10-10
fix(sources): ds-503 accept wider range of hosts values

mirekdlugosz:
Allow user to delimit hosts (hostnames, IP addresses or IP ranges) by space or newline, in addition to comma. This brings new UI in line with old UI, which allowed wider range of user values and transformed them automatically before submitting to backend.

Relates to JIRA: DISCOVERY-503

Questions / points for discussion:

* Should we also have client-side validation of values? Backend rules for what counts as valid host address are [somewhat complex](https://github.com/quipucords/quipucords/blob/main/quipucords/api/source/serializer.py#L276) and I don't really want to repeat them here, but we can definitely fail early for things that are obviously wrong - empty values, invalid host names, multiple hosts where only one should be applied.

* This function is applied when submitting a form and is shared for both "network sources" and "http sources". The problem is, network sources might contain multiple hosts while all the others require exactly one hostname. Should I extend this PR to handle that case somehow? And how?
    1. Take the value of "hosts" field verbatim and send as only member of list. If user types space or comma, this is obviously invalid address that backend will reject
    2. Transform value of "hosts" field, as we do now, but send only first list element. This might lead to what appears to be silent data loss, as typing space or comma is valid, but everything after space or comma is silently dismissed.
    3. ???

### DIFF
--- a/src/helpers/helpers.ts
+++ b/src/helpers/helpers.ts
@@ -99,6 +99,19 @@ const getTimeDisplayHowLongAgo = (timestamp: MomentInput, { devMode = DEV_MODE }
 const formatDate = (date: Date) => moment.utc(date).format('DD MMMM Y, h:mm A z');
 
 /**
+ * Normalizes hosts textarea content into array that can be submitted to backend.
+ *
+ * @param {string} data - host textarea content.
+ * @returns Array of host values which will be submitted to backend.
+ */
+const normalizeHosts = (data?: string) =>
+  data
+    ?.trim()
+    ?.replaceAll(/\\n|\\r|\s/g, ',')
+    ?.split(',')
+    ?.filter(Boolean);
+
+/**
  * Normalizes a total count to a non-negative number less than a given modulus,
  * optionally based on development mode.
  *
@@ -222,6 +235,7 @@ const helpers = {
   getTimeDisplayHowLongAgo,
   getTitleImg,
   formatDate,
+  normalizeHosts,
   normalizeTotal,
   DEV_MODE,
   PROD_MODE,
--- a/src/helpers/__tests__/helpers.test.ts
+++ b/src/helpers/__tests__/helpers.test.ts
@@ -125,6 +125,36 @@ describe('formatDate', () => {
   });
 });
 
+describe('normalizeHosts', () => {
+  it('should accept values separated by space', () => {
+    const input = '127.0.0.1 127.0.0.2';
+    const expected = ['127.0.0.1', '127.0.0.2'];
+
+    expect(helpers.normalizeHosts(input)).toEqual(expected);
+  });
+
+  it('should accept values separated by newline (\\n)', () => {
+    const input = '127.0.0.1\n127.0.0.2';
+    const expected = ['127.0.0.1', '127.0.0.2'];
+
+    expect(helpers.normalizeHosts(input)).toEqual(expected);
+  });
+
+  it('should accept values separated by newline (\\r)', () => {
+    const input = '127.0.0.1\r127.0.0.2';
+    const expected = ['127.0.0.1', '127.0.0.2'];
+
+    expect(helpers.normalizeHosts(input)).toEqual(expected);
+  });
+
+  it('should filter out empty values', () => {
+    const input = '127.0.0.1\n 127.0.0.2 ';
+    const expected = ['127.0.0.1', '127.0.0.2'];
+
+    expect(helpers.normalizeHosts(input)).toEqual(expected);
+  });
+});
+
 describe('normalizeTotal', () => {
   it('should normalize total values', () => {
     expect(helpers.normalizeTotal({ count: 142 }, true, undefined)).toBe(42);
--- a/src/views/sources/__tests__/__snapshots__/addSourceModal.test.tsx.snap
+++ b/src/views/sources/__tests__/__snapshots__/addSourceModal.test.tsx.snap
@@ -5,9 +5,7 @@ exports[`AddSourceModal should call onSubmit with the correct filtered data when
   [
     {
       "credentials": [],
-      "hosts": [
-        "",
-      ],
+      "hosts": [],
       "name": "Test Source",
       "options": {
         "use_paramiko": false,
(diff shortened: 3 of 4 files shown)


src/helpers/helpers.ts
```diff
@@ -90,6 +90,19 @@ const getTimeDisplayHowLongAgo = (timestamp: MomentInput, { devMode = DEV_MODE }
...
+ * Normalizes hosts textarea content into array that can be submitted to backend.
+ *
+ * @param {string} data - host textarea content.
+ * @returns Array of host values which will be submitted to backend.
+ */
+const normalizeHosts = (data: string) => {
+  return data
+    .replaceAll(/\\n|\\r|\s/g, ',')
+    .split(',')
+    .filter(host => !!host || host.length > 0);
```
cdcabrera, line 103: require: filter provides redundant checks any values that come back as an empty string should be `falsey` meaning we could just use `host => !!host);`

...but we've been avoiding the double bang since it's hard to read and it's still a hack that's only made its way back in because of newer js/ts... we've been using...

```
...
    .replaceAll(/\\n|\\r|\s/g, ',')
    .split(',')
    .filter(Boolean);
```

optional
you may also consider replacing `\s` too, and/or using something like the JS string `.trim()` before doing your replace

cdcabrera (changes requested):
Good first iteration. The helper appears to function for filtering against the values in your test. 

Highly recommend you review pr #460 first since it refactors the addSourceModal and this functionality could be added after that work, or parallel, see @nicolearagao 

The [UI `1.8` hosts check was a little bit more in-depth, but could be added independent or completely refactored since it got a little "funky"](https://github.com/quipucords/quipucords-ui/blob/1.8.0/src/components/addSourceWizard/addSourceWizardStepTwo.js#L47).

src/helpers/helpers.ts
mirekdlugosz: Everything done in 7306a293deedf400f2cd47cefb350c8a44a32b7e

mirekdlugosz:
> Highly recommend you review pr #460 first since it refactors the addSourceModal and this functionality could be added after that work, or parallel, see @nicolearagao

We can hold merging this one until #460 is merged. It seems we both touch the same code, so there's going to be a conflict. And it's probably easier to resolve for me, as I change literally single line.

> The [UI `1.8` hosts check was a little bit more in-depth, but could be added independent or completely refactored since it got a little "funky"](https://github.com/quipucords/quipucords-ui/blob/1.8.0/src/components/addSourceWizard/addSourceWizardStepTwo.js#L47).

Yeah, that's first of my questions - the old UI had some field content validation code, which new version does not have at all. I feel we _should_ have it, but I'm looking for input from the rest of the team here. I can add validation as part of this PR, or in new one.

So far I only touched "take user input and transform it to something that backend accepts" part. I believe in old UI it was [addSourceWizardStepTwo.js#L205](https://github.com/quipucords/quipucords-ui/blob/1.8.0/src/components/addSourceWizard/addSourceWizardStepTwo.js#L205), which is fundamentally "split on comma, new line or space". Here we have the same, + removing empty elements.

nicolearagao (commented):
ABout " Should we also have client-side validation of values? Backend rules for what counts as valid host address are [somewhat complex](https://github.com/quipucords/quipucords/blob/main/quipucords/api/source/serializer.py#L276) and I don't really want to repeat them here, but we can definitely fail early for things that are obviously wrong - empty values, invalid host names, multiple hosts where only one should be applied."
I know that 1.8 did a lot more than this, and as per David's comment above, it started to get a little funky. One thing we could do, and I believe it was mentioned by you @mirekdlugosz either in a meeting or in the doc, is not closing the modal when we hit save and there's a problem. I don't know how clear are the messages returned by the backend regarding this specific matter, but at least we could avoid the frustration of having to fill the form again from scratch. After this, we might decide this is still not good enough and improve validations in front-end, but  I would suggest this as first step.

cdcabrera (approved):
In the short term lgtm! Obviously some refactoring may need to take place if we wait for #460

mirekdlugosz:
@nicolearagao yes, I put the idea about not closing the modal into the doc. And while I think this is direction we should be heading, I believe implementation is much more complex than simple field validation I had in my mind. I would prefer to not extend this PR to cover also the idea of not closing the modal.

mirekdlugosz:
@cdcabrera @nicolearagao 5d19d44814cf2d9fd6c1394e7f9815fa51696d14 adds validation rules for `hosts` field. This is a bit tangent, so we can remove it from this PR and I will submit it in new one. But I was under the impression we agree this is the right thing to do.

I started with validation rules from old UI, but they were wrong - they rejected values that backend considers valid. So I mostly copy-pasted regexes used by backend. In general I think UI should be no more strict than backend - it's fine that UI allows value that backend rejects, but it's unacceptable that UI rejects value that backend allows.

One thing I don't like is that this validation is run immediately upon opening modal. So you open the modal and field is marked as invalid. See screenshot below. Technically correct, but poor UX - we shouldn't scream about errors in places that user did not had a chance to touch yet. However, I don't know if there's a way to tell the difference between "this is default empty value" (no validation needed) and "user typed something and then removed it" (sorry user, this is invalid).

I also removed mention of wildcards from helper text, as it was incorrect. `*` and `?` are rejected by server. `[]` is allowed, but in the form `[start:stop]`, not `[abc]`.


Screenshot from 2024-10-14 15-03-54

nicolearagao (approved):
"One thing I don't like is that this validation is run immediately upon opening modal." Even though I don't like this behavior, and I also agree that it's not ideal from UX perspective, IMO not a blocker. @cdcabrera Do you remember how this worked in v1? I took a look at the code and couldn't figure it out what prevemted the same from happening.

mirekdlugosz:
@nicolearagao in old UI fields were marked as invalid only after changing. So when I opened a form, all fields were valid. If I typed something in mandatory field and then removed it, that field was marked as invalid.

mirekdlugosz:
As discussed during standup, I removed validation commit from this PR. I will create new PR for validation only and link to it here.

mirekdlugosz:
Opened https://github.com/quipucords/quipucords-ui/pull/492 as followup.

PR #478, opened 2024-10-10
feat(scans): ds-437 add close button to modal

nicolearagao:
Add close button to last-scanned modal to keep consistency with all other modals. 

### Notes
- fix to [Mirek] “Last Scanned” modal does not have “Close” link/button at the bottom. Every other modal has it (create credential / source, Last connected in Sources, Credentials / Sources modal - but see above). For consistency, that one should have it as well.

1. `$ npm test`

### Local run check
1. `$ npm run start`

## Example

Screenshot from 2024-10-10 19-29-03

## Updates issue/story

### DIFF
--- a/src/views/scans/showScansModal.tsx
+++ b/src/views/scans/showScansModal.tsx
@@ -21,6 +21,7 @@ interface ShowScansModalProps {
   scanJobs?: Pick<ScanJobType, 'id' | 'end_time' | 'report_id' | 'status'>[];
   onDownload?: (number) => void;
   onClose?: () => void;
+  actions?: React.ReactNode[];
 }
 
 const ShowScansModal: React.FC<ShowScansModalProps> = ({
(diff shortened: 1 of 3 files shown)


src/views/scans/showScansModal.tsx
```diff
@@ -88,6 +90,7 @@ const ShowScansModal: React.FC<ShowScansModalProps> = ({
       title={t('view.label', { context: 'scans-names', name: scan?.name })}
       isOpen={isOpen}
       onClose={() => onClose()}
+      {...(actions && { actions })}
```
cdcabrera, line 93: optional: Since we're already handling `onClose` it would be more efficient to leverage our existing work. Since we're not likely to reuse this component, and it's one less setting outside of the modal itself... 

```
actions={[
        <Button key="close" variant="secondary" onClick={() => onClose()}>
          {t('table.label', { context: 'close' })}
        </Button>
      ]}
```
or if you want to still expose the actions with an internal default... 

```
actions={
        actions || [
          <Button key="close" variant="secondary" onClick={() => onClose()}>
            {t('table.label', { context: 'close' })}
          </Button>
        ]
      }
```

cdcabrera (approved):
What you have works! 

If you want to minimally optimize, I recommend leveraging the same behavior as the onClose... see optional recommendation

mirekdlugosz (approved):
I only looked at screenshot and code, and they appear good to me!

PR #485, opened 2024-10-14
fix: ds-438 add refetch interval to query helper

nicolearagao:
Add the parameter refetchInterval provided by react-query and set it to one minute. 

### Notes
- fix to [Mirek] Scans page refreshes automatically every 15 minutes. That seems a bit high - every 5 minutes would probably be enough.
2024-10-02 Is this a timer issue, or is this a browser issue? The timer should be 1 minute to be consistent with the old UI. Mirek suggests that in some cases the browser may pause background tabs.
If the issue is in our code having the “wait” number too high, please lower the number.
Future idea: configurable refresh time.
[Mirek] 2024-10-09 I found this JIRA that might be related: [DISCOVERY-443](https://issues.redhat.com/browse/DISCOVERY-443)
- This doesn't solve the issue raised on DISCOVERY-443, we can always lower the interval in whch we refresh the page for scans, but according to the acceptance criteria this needs to happen when a few pre-requisites are met. The solution for this problem requires a more complex implementation, therefore I would recommend  attempting to solve this in a second iteration.

1. `$ npm test`

### Local run check
1. `$ npm run start`

## Updates issue/story


cdcabrera (changes requested):
it would be good to leverage the left-over dotenv parameter here instead of hardcoding this value
- https://github.com/quipucords/quipucords-ui/blob/main/.env#L17

.env
```diff
@@ -14,7 +14,7 @@ REACT_APP_AUTH_COOKIE=user
 REACT_APP_AUTH_COOKIE_EXPIRES=0.5
 REACT_APP_AJAX_TIMEOUT=60000
 REACT_APP_TOAST_NOTIFICATIONS_TIMEOUT=8000
-REACT_APP_POLL_INTERVAL=120000
+REACT_APP_POLL_INTERVAL=60000
```
cdcabrera, line 17: comment/optional: apologies I missed the statement from @mirekdlugosz that the old timer was 1 minute. 

the old timer was this setting `120000` we updated a couple of years back, [see the blame](https://github.com/quipucords/quipucords-ui/blame/1.8.0/.env#L15). Whether it was working correctly that's a parallel issue. But if we're trying to mirror the old behavior we should just leave the value as is... if we want more refreshes I'm not necessarily opposed

src/helpers/queryHelpers.ts
```diff
@@ -60,10 +60,13 @@ export const useServiceQuery = <TItem, TColumnKey extends string, TSortableColum
   columnOrderMap?: Record<TSortableColumnKey, string>;
   tableState: TableState<TItem, TColumnKey, TSortableColumnKey>;
   setRefreshTime?: (date: Date) => void;
-}) =>
-  useQuery<ServiceQueryResult<TItem>>({
+}) => {
+  const pollInterval = Number(process.env.REACT_APP_POLL_INTERVAL) || 60000;
```
cdcabrera, line 64: required:
we should move to the specific method instead of the generic handling of `Number`. what you have works but can lead to subtle JS behavior alterations. 

```
Number.parseInt(process.env.REACT_APP_POLL_INTERVAL, 10)
```

optional:
instead of a hardcoded value we could also make the value `undefined`. It'll default towards whatever was there originally. and coincidentally would restore the exact same handling of this value as `v1.8` ... https://github.com/quipucords/quipucords-ui/blob/1.8.0/src/common/helpers.js#L436

Plus it's one less thing to track down for future us

```
Number.parseInt(process.env.REACT_APP_POLL_INTERVAL, 10) || undefined
```

cdcabrera (changes requested):
close! one change, comment about legacy behavior, and an optional

.env
nicolearagao: I believe it was a consensus that 1 minute poll intervals was ideal. I will keep REACT_APP_POLL_INTERVAL=60000

src/helpers/queryHelpers.ts
nicolearagao: This didn't really work, because Number.parseInt expects a string, so it complains beause it might be undefined. 
`  Number.parseInt(process.env.REACT_APP_POLL_INTERVAL, 10) || undefined`
As an alternative I check it it exists first, then try to parse it, what do ya think ?
```
  const pollInterval = process.env.REACT_APP_POLL_INTERVAL 
  ? Number.parseInt(process.env.REACT_APP_POLL_INTERVAL, 10) 
  : undefined;
```

src/helpers/queryHelpers.ts
```diff
@@ -60,10 +60,15 @@ export const useServiceQuery = <TItem, TColumnKey extends string, TSortableColum
   columnOrderMap?: Record<TSortableColumnKey, string>;
   tableState: TableState<TItem, TColumnKey, TSortableColumnKey>;
   setRefreshTime?: (date: Date) => void;
-}) =>
-  useQuery<ServiceQueryResult<TItem>>({
+}) => {
+  const pollInterval = process.env.REACT_APP_POLL_INTERVAL
+    ? Number.parseInt(process.env.REACT_APP_POLL_INTERVAL, 10)
+    : undefined;
```
cdcabrera, line 66: optional: 
some more subtle js behavior this also returns undefined... this could be simplified and we can leverage the same property name 
```
const refetchInterval = process.env.REACT_APP_POLL_INTERVAL && Number.parseInt(process.env.REACT_APP_POLL_INTERVAL, 10);

...
refetchOnWindowFocus: !helpers.DEV_MODE,
refetchInterval,
...
```

cdcabrera (approved):
optional recommended clean up work... otherwise works!

cdcabrera (approved):
github is acting odd, my 2nd approval ha

PR #486, opened 2024-10-16
refactor: ds-438 display items alphabetical order

nicolearagao:
Display list of sources and credentials in alphabetical order

### Notes
- fix to [Mirek] A list of Sources used by a credential, or sources used by a scan (the one that appears in modal) is sorted by related items id. It would be more natural to sort it alphabetically.
Not tested, but I assume the same applies to list of credentials used by a source

1. `$ npm test`

### Local run check
1. `$ npm run start`

## Example

Screenshot from 2024-10-15 21-38-32
Screenshot from 2024-10-15 21-38-25
Screenshot from 2024-10-15 21-38-16

## Updates issue/story

### DIFF
--- a/src/views/credentials/viewCredentialsList.tsx
+++ b/src/views/credentials/viewCredentialsList.tsx
@@ -303,9 +303,11 @@ const CredentialsListView: React.FunctionComponent = () => {
         ]}
       >
         <List isPlain isBordered>
-          {sourcesSelected.map(c => (
-            <ListItem key={c.name}>{c.name}</ListItem>
-          ))}
+          {sourcesSelected
+            .sort((a, b) => a.name.localeCompare(b.name))
+            .map(c => (
+              <ListItem key={c.name}>{c.name}</ListItem>
+            ))}
         </List>
         {/* TODO: his modal should go on a list of getting it's own component * check PR #381 for details */}
       </Modal>
--- a/src/views/scans/viewScansList.tsx
+++ b/src/views/scans/viewScansList.tsx
@@ -273,7 +273,9 @@ const ScansListView: React.FunctionComponent = () => {
         ]}
       >
         <List isPlain isBordered>
-          {scanSelectedForSources?.sources.map(s => <ListItem key={s.id}>{s.name}</ListItem>)}
+          {scanSelectedForSources?.sources
+            .sort((a, b) => a.name.localeCompare(b.name))
+            .map(s => <ListItem key={s.id}>{s.name}</ListItem>)}
         </List>
       </Modal>
       <ShowScansModal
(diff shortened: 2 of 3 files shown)


mirekdlugosz (approved):

src/views/credentials/viewCredentialsList.tsx
```diff
@@ -303,9 +303,11 @@ const CredentialsListView: React.FunctionComponent = () => {
         ]}
       >
         <List isPlain isBordered>
-          {sourcesSelected.map(c => (
-            <ListItem key={c.name}>{c.name}</ListItem>
-          ))}
+          {sourcesSelected
+            .sort((a, b) => a.name.localeCompare(b.name))
```
infinitewarp, line 307: Would you expect the results to be sorted already in the API's response so you don't have to sort in the UI code here? We can implement that on the server if you like, and I think it should be relatively easy.

The main API responses should already be sortable by sending the `ordering=name` GET query param. For example:

```
http -b get 127.0.0.1:8000/api/v2/credentials/ ordering==name page_size==3 Authorization:"Token $API_TOKEN" | jq '.results[] | {id, name}'
{
  "id": 10011,
  "name": "ansible-admin"
}
{
  "id": 27796,
  "name": "credv2"
}
{
  "id": 10008,
  "name": "net-dev01-baremetalqpc"
}
```

However, if this list is coming from the _nested_ object data within another object, like the list of sources that is returned as part of each credential object, then the API does not currently provide any option to sort that sub-object. I confirmed that currently the API doesn't sort the objects in the sources attribute of a credential. We _could_ bake in a better default behavior, though. Example of current unsorted output:

```
❯ http -b get 127.0.0.1:8000/api/v2/credentials/10008/ Authorization:"Token $API_TOKEN" | jq '{id, sources}'
{
  "id": 10008,
  "sources": [
    {
      "id": 19626,
      "name": "net-dev01-8-mixed-good-bad"
    },
    {
      "id": 19628,
      "name": "net-dev01-2-good-x86_64"
    },
    {
      "id": 19625,
      "name": "net-dev01-3-good-1-arm64-2-x86_64"
    },
    {
      "id": 29905,
      "name": "net-several"
    }
  ]
}
```

I'm not saying to abandon the work you've started here, because this looks totally fine to me. Just asking for opinions on whether you think there's value in making the server _also_ behave with better sorted items here.

infinitewarp (approved):

src/views/credentials/viewCredentialsList.tsx
nicolearagao: @infinitewarp I personally don't see extra benefits in changing the api, since the solution in the UI is fairly simple. Regarding expectations of the results received, I am not sure. I don't have a lot of experience to say that this is the pattern adopted everywhere, so maybe @cdcabrera could answer better.

bruno-fs: this only works if you have a single page of results.

Sorting should be done through the database, not from the limited information a client can access.

PR #488, opened 2024-10-16
feat(aboutModal): ds-808 display user, status info

cdcabrera:
## What's included

- feat(aboutModal): ds-808 display user, status info
   * aboutModal, display user and status info
   * viewLayout, use title image helper
   * viewLayoutToolbar, add about modal open, close
   * helpers, add ui_version, getCurrentDate, getTitleImg
   * useStatusApi, use status api response

### Notes
- modifies the brand image into a centralized helper to avoid recreating the logic in multiple places
- adds the centralized `getCurrentDate` and `UI_VERSION` helpers from `v1.8`
- what this doesn't do (these items can be iterated on if necessary)
   - ~skips adding the UI browser type from `v1.8` since it involves adding in one more NPM.~
   - skips the `v1.8` custom copy/paste behavior.
- running locally
   - the randomized data local run using `$ npm start` (generated from the `spec` file) returns a `500` level error from the generated `status` call, unclear if there is something malformed in the `spec` since it complains about the shape of the response
   - there's a slight delay in the normal/`staging` run when the initial `status` api call is made. this may, or may not happen, in the standard run. to compensate we use a quick css transition to mask the load jump. if it becomes obvious we may need another solution to indicate loading for `api_version` and `server_version`
- The use of `Red Hat Inc` in the copyright/trademark... is pulled from `v1.8`, confirmation may be needed on the use of `Inc`

### Local run check
1. `$ npm run start:stage`
1. select the question mark circle next to the user name
   - confirm it displays the about modal similar to the example

## Example

_*note: the ui version 1.8.0 is currently the package.json version. this is local run behavior. the format for the ui version in production follows the format `0.0.0.XXXXXXX`, a combination of the package.json version and git hash_

Screenshot 2024-10-15 at 9 48 44 PM

## Updates issue/story

[DISCOVERY-808](https://issues.redhat.com/browse/DISCOVERY-808)

### DIFF
--- a/src/components/aboutModal/aboutModal.tsx
+++ b/src/components/aboutModal/aboutModal.tsx
@@ -1,0 +1,102 @@
+import React, { useEffect, useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import { AboutModal as PfAboutModal, TextContent, TextList, TextListItem } from '@patternfly/react-core';
+import { detect } from 'detect-browser';
+import moment from 'moment/moment';
+import { helpers } from '../../helpers';
+import { useUserApi } from '../../hooks/useLoginApi';
+import { useStatusApi, type ApiStatusSuccessType } from '../../hooks/useStatusApi';
+import backgroundImageSrc from '../../images/aboutBg.png';
+
+interface AboutModalProps {
+  currentYear?: string;
+  isOpen?: boolean;
+  onClose?: () => void;
+  titleImg?: string;
+  uiName?: string;
+  uiVersion?: string;
+  useStatus?: typeof useStatusApi;
+  useUser?: typeof useUserApi;
+}
+
+const AboutModal: React.FC<AboutModalProps> = ({
+  currentYear = moment.utc(helpers.getCurrentDate()).format('YYYY'),
+  isOpen = false,
+  onClose = Function.prototype,
+  titleImg = helpers.getTitleImg(),
+  uiName = helpers.UI_NAME,
+  uiVersion = helpers.UI_VERSION,
+  useStatus = useStatusApi,
+  useUser = useUserApi
+}) => {
+  const { t } = useTranslation();
+  const { getStatus } = useStatus();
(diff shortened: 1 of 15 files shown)


src/components/aboutModal/aboutModal.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+  useEffect(() => {
+    if (isOpen && !stats) {
+      getUser().then(username => setUserName(username));
+      getStatus().then(
+        data => setStats(data),
+        error => console.error(`About status error: ${error} `)
+      );
+    }
+  }, [isOpen, getStatus, getUser, stats]);
```
nicolearagao, line 48: Do we need to have stats in the dependency array here ? It's already used in the condition so I had the impression it didn't need to be in the array as well.

src/components/aboutModal/aboutModal.tsx
```diff
@@ -0,0 +1,102 @@
...
+  onClose?: () => void;
+  titleImg?: string;
+  uiName?: string;
+  uiVersion?: string;
+  useStatus?: typeof useStatusApi;
+  useUser?: typeof useUserApi;
+}
+
+const AboutModal: React.FC<AboutModalProps> = ({
+  currentYear = moment.utc(helpers.getCurrentDate()).format('YYYY'),
```
nicolearagao, line 23: should we move from moment lib since now it is considered a legacy project ? I am not sure what we should use instead, but I thought it was worth it to bring the idea here. From a fast search online libs date-fns or dayjs pop up, but I also don't know if supported to downstream builds.

nicolearagao (approved):

infinitewarp (approved):

src/components/aboutModal/aboutModal.tsx
cdcabrera: good catch! you are correct we technically don't need it since it has the potential to create a doom loop... i'll remove it and eslint disable, currently, line `48`

PR #489, opened 2024-10-16
fix: ds-437 remove untreated errors

nicolearagao:
Remove code that throws untreated errors. 

### Notes
- fix to  [Bruno] On the "UI should gracefully handle errors" topic, there are 2 points on the code throwing errors that might be untreated. I don't understand their context, but one of them even has a comment suggesting returning a default value as an alternative.
https://github.com/quipucords/quipucords-ui/blob/cc533bc1944c017569f1246a207c67a442b2b05c/src/helpers/queryHelpers.ts#L76
https://github.com/quipucords/quipucords-ui/blob/cc533bc1944c017569f1246a207c67a442b2b05c/src/hooks/useScanApi.ts#L131

1. `$ npm test`

### Local run check
1. `$ npm run start`

## Updates issue/story


src/hooks/__tests__/useScanApi.test.ts
```diff
@@ -400,6 +400,14 @@ describe('useRunScanApi', () => {
 
     expect(mockOnAddAlert.mock.calls).toMatchSnapshot('callbackError');
   });
+
+  it('should handle missing scan ID and trigger custom error', async () => {
+    const { runScans } = hookResult;
+    mockCreateScan.mockResolvedValue({});
```
cdcabrera, line 406: nice reuse!

cdcabrera (changes requested):
minor changes to alleviate typings

src/helpers/queryHelpers.ts
```diff
@@ -72,8 +72,8 @@ export const useServiceQuery = <TItem, TColumnKey extends string, TSortableColum
         setRefreshTime?.(new Date());
         return response.data;
       } catch (error) {
-        console.error(error);
-        throw error; // You can choose to throw the error or return a default value here
+        console.log('Error during data fetching:', error);
```
mirekdlugosz, line 75: Just an observation: `console.error()` is perfectly valid and supported by browsers.

[Camayoc has a special handler that will look for error messages](https://github.com/quipucords/camayoc/blob/main/camayoc/ui/client.py#L95) and consider a test failing if there are any. So moving that back to `console.error()` has a potential of failing the pipeline. But I think that's a good thing? As far as I can tell, we have not had any failures caused by `console.error()` here, so if it ever happens then probably this is something we should take a look at.

cdcabrera: i agree with the leveraging the `console.error` here. 

we do leverage `error` in a few places under the api calls but we also wrap some of them with a`!helpers.TEST_MODE` check. But this flag should only affect unit testing, not e2e since its running everything... maybe

cdcabrera (approved):
lgtm, nothing blocking...

i would consider @mirekdlugosz `console.error` comment unless you all have conferred off to the side. it could also be something we iterate on, good to know the pipeline will catch them.

src/helpers/queryHelpers.ts
nicolearagao: I will revert it back to console.error and let's see if it affects the pipeline at all.

PR #172, opened 2022-09-27
refactor(viewToolbar): discovery-151 pf4 toolbar

cdcabrera:
## What's included

...

## Example

...

PR #491, opened 2024-10-17
fix: ds-441 use correct API field when sorting scans

mirekdlugosz:
[Scans sorting field is called `most_recent_scanjob__start_time`](https://github.com/quipucords/quipucords/blob/main/quipucords/api/scan/view.py#L261). `most_recent_connect_scan__start_time` is available only in sources API.

We even have that string in constants, but it is unused. I looked into git history and it seems that this mistake was introduced back during PF4 migration. Probably a mis-click when selecting a value from intellisense list.

I tested locally that with that patch, clicking headers in scans list changes the order of items. Let me know if there are more places I should change.

### DIFF
--- a/src/views/scans/useScansQuery.ts
+++ b/src/views/scans/useScansQuery.ts
@@ -29,7 +29,7 @@ export const useScansQuery = ({
     baseUrl: process.env.REACT_APP_SCANS_SERVICE,
     columnOrderMap: {
       name: 'name',
-      most_recent: 'most_recent_connect_scan__start_time'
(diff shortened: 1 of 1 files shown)


cdcabrera (approved):
Does appear to work in local run! Good catch!

If anyone is interested the change popped up in this story [DISCOVERY-151](https://issues.redhat.com/browse/DISCOVERY-151) under pr #172 `v1.0.0`

nicolearagao (approved):
thanks!

PR #492, opened 2024-10-17
feat(sources): ds-503 add hosts field validation

mirekdlugosz:
This was originally part of #475 .

Add function that validates the content of hosts field. It is generally meant to reject empty lists and lists that have too many items (everything but network works against single host), but it also looks into individual hosts and rejects ones that won't be accepted by backend anyway.

Regular expressions used in `hostValid` are [copied from backend](https://github.com/quipucords/quipucords/blob/main/quipucords/api/source/serializer.py#L329).

One thing I don't like in current implementation is that field is marked as invalid right as you open the modal. It should be marked as such only if user touched a field / tries to submit form without touching a field. Hints how to solve that are most appreciated!

### DIFF
--- a/src/helpers/__tests__/helpers.test.ts
+++ b/src/helpers/__tests__/helpers.test.ts
@@ -5,6 +5,7 @@
  * checks.
  */
 import React from 'react';
+import { ValidatedOptions } from '@patternfly/react-core';
 import moment from 'moment';
 import { helpers } from '../helpers';
 
@@ -152,6 +153,46 @@ describe('normalizeHosts', () => {
     const expected = ['127.0.0.1', '127.0.0.2'];
 
     expect(helpers.normalizeHosts(input)).toEqual(expected);
+  });
+});
+
+describe('validateHosts', () => {
+  // values copied from backend test
+  // quipucords/tests/api/source/test_source.py::TestSource::test_create_valid_hosts
+  it.each([
+    ['10.10.181.9'],
+    ['10.10.181.9/16'],
+    ['10.10.128.[1:25]'],
+    ['10.10.[1:20].25'],
+    ['10.10.[1:20].[1:25]'],
+    ['localhost'],
+    ['my_cool_underscore.com'],
+    ['bgimages.com'],
+    ['my_rhel[a:d].company.com'],
+    ['my_rhel[120:400].company.com'],
+    ['my-rhel[a:d].company.com'],
+    ['my-rhel[120:400].company.com'],
+    ['my-rh_el[120:400].comp_a-ny.com']
+  ])('should accept valid host [%s]', host => {
+    expect(helpers.validateHosts(host, Infinity)).toBe(ValidatedOptions.default);
+  });
+
+  it('should reject empty hosts', () => {
+    const input = ',,';
+
+    expect(helpers.validateHosts(input, Infinity)).toBe(ValidatedOptions.error);
+  });
+
+  it('should reject multiple hosts when only one is allowed', () => {
+    const input = '127.0.0.1 127.0.0.2';
+
+    expect(helpers.validateHosts(input, 1)).toBe(ValidatedOptions.error);
+  });
+
+  it('should reject too long host', () => {
+    const input = 'a'.repeat(400);
+
+    expect(helpers.validateHosts(input, Infinity)).toBe(ValidatedOptions.error);
   });
 });
 
--- a/src/helpers/helpers.ts
+++ b/src/helpers/helpers.ts
@@ -6,6 +6,7 @@
  * @module helpers
  */
 import React from 'react';
+import { ValidatedOptions } from '@patternfly/react-core';
 import moment, { type MomentInput } from 'moment';
 import titleImg from '../images/title.svg';
 import titleImgBrand from '../images/titleBrand.svg';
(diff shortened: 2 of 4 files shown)


src/helpers/helpers.ts
```diff
@@ -111,6 +112,50 @@ const normalizeHosts = (data?: string) =>
...
     ?.filter(Boolean);
 
+/**
+ * Validates hosts textarea content.
+ *
+ * @param {string} data - host textarea content.
+ * @param {number} maxHosts - maximum number of hosts that the field should allow.
+ * @returns ValidatedOptions
+ */
+const validateHosts = (data: string | undefined, maxHosts: number) => {
```
cdcabrera, line 122: required: lets use a default for `maxHosts`. reason... old skool js, secondary params are associated with support settings and we can provide a tiny bit of developer "ease of use"

up to you whether you want to use `maxHosts: number = Infinity` or `maxHosts: number = 1` (i partially assume it would default to Infinity)

once you do this your typescript can also be updated towards (and of course the call where you declare `Infinity` in `addSourceModal` can be updated)

```
const validateHosts = (data?: string, maxHosts: number = Infinity) => {
```

src/helpers/helpers.ts
```diff
@@ -111,6 +112,50 @@ const normalizeHosts = (data?: string) =>
...
+    return allRegexes.some(re => new RegExp(re).test(value));
+  };
+
+  if (data === undefined) {
+    return ValidatedOptions.default;
+  }
+
+  const normalizedHosts = normalizeHosts(data);
+
+  if (!normalizedHosts!.length) {
```
cdcabrera, line 144: looks like we forgot the eslint rule `'@typescript-eslint/no-non-null-assertion'` that's on me... apologies. it's difficult to read and there's the baked in JS optional chaining

**required:** all instances of this "non-null-assertion", `!.` should be replaced... prefer optional chaining, `?.`, instead. that would make this line

```
if (!normalizedHosts?.length)
```
and of course update the remaining instances of this

**optional**
add this rule to the eslint config to avoid this scenario again
```
'@typescript-eslint/no-non-null-assertion': 2,
```

src/helpers/__tests__/helpers.test.ts
```diff
@@ -155,6 +156,46 @@ describe('normalizeHosts', () => {
...
+    ['10.10.[1:20].[1:25]'],
+    ['localhost'],
+    ['my_cool_underscore.com'],
+    ['bgimages.com'],
+    ['my_rhel[a:d].company.com'],
+    ['my_rhel[120:400].company.com'],
+    ['my-rhel[a:d].company.com'],
+    ['my-rhel[120:400].company.com'],
+    ['my-rh_el[120:400].comp_a-ny.com']
+  ])('should accept valid host [%s]', host => {
```
cdcabrera, line 176: sweet use! good call!

cdcabrera (changes requested):
Almost there! Some old Typescript we should squash

src/helpers/helpers.ts
mirekdlugosz: Done c8a96d131f94ebd901a405ef37d2ca19ebca73a5 .

I still supply both arguments on caller side, to make it more explicit that there are different rules based on source type.

src/helpers/helpers.ts
mirekdlugosz: Done in c8a96d131f94ebd901a405ef37d2ca19ebca73a5 , along with optional.

I didn't like this `variable!`, and felt it is not necessary. In cbfbcd6faa3b010f2204e122283e436d06a328ee I took another look and modified `normalizeHosts` in a way that makes both `?.` and `!.` not necessary.

This is another commit so we can easily drop it if there are any problems with that approach.

src/helpers/__tests__/helpers.test.ts
mirekdlugosz: Thanks :smiley:

mirekdlugosz:
On top of fixes for things pointed out in review, I introduced a bit larger change in cbfbcd6faa3b010f2204e122283e436d06a328ee .

It ensures that `normalizeHosts` returns array, not array or undefined. That makes things a bit more readable in `validateHosts`, but the cost is that it is no longer pure arrow function. Not sure if that matters.

I also changed default SourceFormType state to have `hosts` as `undefined` (instead of empty string). This allows us to differentiate between "field is empty because modal has been just opened" (field not marked as invalid) and "field is empty because user typed something and removed it" (field marked as invalid). `normalizeHosts` still changes that to empty list, and we could prevent form submission once we have that code.

However, I am not sure if that `undefined` will not cause any issues elsewhere, so I am open to dropping this commit.

cdcabrera (approved):
tested using similar regular expressions used in unit tests. appears to function as intended

>there was some history around the specific old limits setup for subnets... unclear if that's changed or because of the openness of the new checks/currently missing validation, if that's still a concern... regardless it can be iterated on
>
> - https://github.com/quipucords/quipucords-ui/blob/1.8.0/src/components/addSourceWizard/__tests__/addSourceWizardStepTwo.test.js#L98-L116

mirekdlugosz:
In c8baf9da1c682f660d4831cd6aaabeca837e9e02 I added some strings from old unit tests. I skipped following:

* `0.0.0.0:22` this is indeed invalid, but gets through new validation and I don't think I can catch it without making that code much more complex. Since the long term goal is to perform all validation on server side, I think we can ignore it for now.
* `0.0.0.12/15` this used to be considered invalid, but I have no idea why. This is perfectly valid [CIDR block](https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing#IPv4_CIDR_blocks). It will result in a scan of over 130 000 hosts, which is likely not what anyone needs, but technically it is correct and I don't think we can know better than the user what they want.
* `0.0.0.12/33` this is indeed invalid, but again - we would have to use more complex code to catch it. For now we allow any two-digit number.

PR #493, opened 2024-10-17
Fix/scans

nicolearagao:
Populate Scan time field with start_time if end_time is not available.
Sort scanjobs to be displayed latest first in descending order

### Notes
- fix to  [Mirek] “Last Scanned” modal. While scan is still running, “Scan time” is empty. Right now it seems to refer to end_time field, which is not present in API response for a running scan. It should either always display start_time field, or display end_time if present, but fall back to start_time
- Scans [Mirek] “Last Scanned” modal does not support /api/v1/scans/<id>/jobs/ paging. It will only display first page of results, and results are sorted from oldest to newest. “X scans have run” above table is populated with a length of results, not response count field value, so there is no way of knowing that the most recent scan results are not available through UI.

1. `$ npm test`

### Local run check
1. `$ npm run start`

## Example

Screenshot from 2024-10-17 17-14-57

## Updates issue/story

### DIFF
--- a/src/views/scans/showScansModal.tsx
+++ b/src/views/scans/showScansModal.tsx
@@ -18,7 +18,7 @@ import { type Scan, type ScanJobType } from '../../types/types';
 interface ShowScansModalProps {
   isOpen: boolean;
   scan?: Pick<Scan, 'name'>;
-  scanJobs?: Pick<ScanJobType, 'id' | 'end_time' | 'report_id' | 'status'>[];
+  scanJobs?: Pick<ScanJobType, 'id' | 'start_time' | 'report_id' | 'status' | 'end_time'>[];
   onDownload?: (report_id: number) => void;
   onClose?: () => void;
   actions?: React.ReactNode[];
@@ -35,7 +35,6 @@ const ShowScansModal: React.FC<ShowScansModalProps> = ({
   const { t } = useTranslation();
   const [activeSortIndex, setActiveSortIndex] = useState<number | undefined>();
   const [activeSortDirection, setActiveSortDirection] = useState<'asc' | 'desc' | undefined>();
-  // const [selectedJobs, setSelectedJobs] = useState<number[]>([]);
 
   const getSortParams = (columnIndex: number): ThProps['sort'] => ({
     sortBy: {
@@ -49,40 +48,6 @@ const ShowScansModal: React.FC<ShowScansModalProps> = ({
     },
     columnIndex
   });
-
-  /* ToDo: This code is for future use around Discovery-437. If it isn't used when the issue has resolved remove it
- *
-  const isJobSelected = (job: ScanJobType) => { return selectedJobs.includes(job.id); };
-  const isJobSelectable = (job: ScanJobType) => { return true || job.status === "completed"; };
-  const selectableJobs = scanJobs ? scanJobs.filter(isJobSelectable) : [];
-  const onSelectJob = (job: ScanJobType, rowIndex: number, isSelecting: boolean) => {
-      setSelectedJobs((prevSelected) => {
-          const otherSelected = prevSelected.filter((j) => j !== job.id);
-          return isSelecting && isJobSelectable(job) ? [...otherSelected, job.id] : otherSelected;
-        });
-  };
-  const areAllJobsSelected = false;
-
-  const selectAllJobs = (isSelecting: boolean = true) => {
-      setSelectedJobs(isSelecting ? selectableJobs.map((j) => j.id) : []);
-  };
-  const getSortableRowValues = (scanJob: ScanJobType): (Date | string)[] => {
-  const { end_time, status } = scanJob;
-  return [end_time, status];
-};
-
-  let sortedJobs = scanJobs;
-  if (activeSortIndex !== undefined && scanJobs) {
-    sortedJobs = scanJobs.sort((a, b) => {
-      const aValue = getSortableRowValues(a)[activeSortIndex];
-      const bValue = getSortableRowValues(b)[activeSortIndex];
-      console.log({ aValue, bValue });
-      if (activeSortDirection === 'asc') {
-        return (aValue as string)?.localeCompare(bValue as string);
-      }
-      return (bValue as string)?.localeCompare(aValue as string);
-    });
-  }*/
 
   return (
     <Modal
@@ -101,12 +66,6 @@ const ShowScansModal: React.FC<ShowScansModalProps> = ({
           <Table aria-label="Scan jobs table" ouiaId="scan_jobs_table">
             <Thead>
               <Tr>
-                {/* <Th
-                                    select={{
-                                        onSelect: (_event, isSelecting) => selectAllJobs(isSelecting),
-                                        isSelected: areAllJobsSelected
-                                    }}
-                                /> */}
                 <Th aria-labelledby="Sort by column" sort={getSortParams(0)}>
                   Scan Time
                 </Th>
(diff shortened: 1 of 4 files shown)


mirekdlugosz:
Scans that are still running have "Scan time" column in modal filled.

If there are multiple rescans of a single scan, in modal the most recent one will be at the top. However, sorting is applied locally.

I am confused by this part of description:

> * Scans [Mirek] “Last Scanned” modal does not support /api/v1/scans//jobs/ paging. It will only display first page of results, and results are sorted from oldest to newest. “X scans have run” above table is populated with a length of results, not response count field value, so there is no way of knowing that the most recent scan results are not available through UI.

As far as I can see in the PR, nothing about paging is changed? I have not tried it, but I believe the table in modal will still display only single page of results, won't it?

Which also means that the scan appearing on top might not be the most recent scan - it will be the most recent scan from first page of results. If there happens to be newer one on second or later pages, that one will not appear.

(As discussed elsewhere, default paging is 100 items, so most users will not be affected by that yet; I am only bringing all that up because PR description mentions it)

src/views/scans/showScansModal.tsx
```diff
@@ -18,12 +18,22 @@ import { type Scan, type ScanJobType } from '../../types/types';
 interface ShowScansModalProps {
   isOpen: boolean;
   scan?: Pick<Scan, 'name'>;
-  scanJobs?: Pick<ScanJobType, 'id' | 'end_time' | 'report_id' | 'status'>[];
+  scanJobs?: (Pick<ScanJobType, 'id' | 'start_time' | 'report_id' | 'status'> & { end_time: Date | null })[];
```
cdcabrera, line 21: required: `end_time` is already included in `ScanJobType`. if there is the potential that `end_time` doesn't exist it should be made optional within the the `ScanJobType` under `types.ts`

Change `ScanJobType` under `types.ts`
```
...
  report_id: number;
  start_time: Date;
  end_time?: Date;
  systems_count: number;
...
```

Change this current line towards
```
scanJobs?: Pick<ScanJobType, 'id' | 'start_time' | 'report_id' | 'status' | 'end_time'>[];
```

cdcabrera (changes requested):
typings don't line up...

nicolearagao:
> Scans that are still running have "Scan time" column in modal filled.
> 
> If there are multiple rescans of a single scan, in modal the most recent one will be at the top. However, sorting is applied locally.
> 
> I am confused by this part of description:
> 
> > * Scans [Mirek] “Last Scanned” modal does not support /api/v1/scans//jobs/ paging. It will only display first page of results, and results are sorted from oldest to newest. “X scans have run” above table is populated with a length of results, not response count field value, so there is no way of knowing that the most recent scan results are not available through UI.
> 
> As far as I can see in the PR, nothing about paging is changed? I have not tried it, but I believe the table in modal will still display only single page of results, won't it?
> 
> Which also means that the scan appearing on top might not be the most recent scan - it will be the most recent scan from first page of results. If there happens to be newer one on second or later pages, that one will not appear.
> 
> (As discussed elsewhere, default paging is 100 items, so most users will not be affected by that yet; I am only bringing all that up because PR description mentions it)

@mirekdlugosz I see how this is confusing. Your understanding is correct. No changes related to pagination were applied to this PR. We agreed that for this issue specifically, we would make sure the latest was on top, pagination would be fixed in the near future. I copied and pasted the description of the problem from the docs, wthout explaining what exactly was being fixed, my apologies.

mirekdlugosz (approved):
@nicolearagao that's fine. Glad we are on the same page. We only need to make sure that we don't mark paging as fixed in the doc - let's keep it open there, we'll get to it eventually.

Other changes look good :+1:

src/views/scans/showScansModal.tsx
```diff
@@ -152,4 +110,4 @@ const ShowScansModal: React.FC<ShowScansModalProps> = ({
   );
 };
 
-export { ShowScansModal as default, ShowScansModal, type ShowScansModalProps };
+export { ShowScansModal, type ShowScansModalProps };
```
cdcabrera, line 113: required: restore the `default`

src/views/scans/showScansModal.tsx
```diff
@@ -50,39 +49,12 @@ const ShowScansModal: React.FC<ShowScansModalProps> = ({
...
-      }
-      return (bValue as string)?.localeCompare(aValue as string);
-    });
-  }*/
+  // Sort scanJobs by scan time (latest first)
+  const sortedScanJobs = (scanJobs || []).sort((a, b) => {
+    const timeA = a.end_time ? new Date(a.end_time).getTime() : new Date(a.start_time).getTime();
+    const timeB = b.end_time ? new Date(b.end_time).getTime() : new Date(b.start_time).getTime();
+    return timeB - timeA; // Sort in descending order (latest first)
+  });
```
cdcabrera, line 57: optional: this entire helper can be made similar to how we handled the other inline where we know `start_time` exists

```
  const sortedScanJobs = (scanJobs || []).sort((a, b) => {
    const timeA = new Date(a.end_time || a.start_time).getTime();
    const timeB = new Date(b.end_time || b.start_time).getTime();
    return timeB - timeA; // Sort in descending order (latest first)
  });
```

future:
we should review the `scanJobs || []` as a default param instead

ruda (approved):
Tested locally with an OCP scan, LGTM.

src/views/scans/showScansModal.tsx
```diff
@@ -92,21 +64,15 @@ const ShowScansModal: React.FC<ShowScansModalProps> = ({
...
           <Table aria-label="Scan jobs table" ouiaId="scan_jobs_table">
             <Thead>
               <Tr>
-                {/* <Th
-                                    select={{
-                                        onSelect: (_event, isSelecting) => selectAllJobs(isSelecting),
-                                        isSelected: areAllJobsSelected
-                                    }}
-                                /> */}
                 <Th aria-labelledby="Sort by column" sort={getSortParams(0)}>
```
mirekdlugosz, line 76: Since we are introducing local sorting mechanism, and the table was not properly hooked to API requests for sorting, would you mind removing `sort` argument here?

```suggestion
                <Th aria-labelledby="Sort by column">
```

The same goes to line 79 (Scan Result header).

This will tie to the following issue from the doc:

> “Last Scanned” modal. Clicking table header items inside modal does not actually change the order of items in table.

Not fully a solution, but at least we won't present UI elements that wouldn't do anything.

nicolearagao: Mirek, I will introduce this change in a followup PR.

nicolearagao: Mirek, I will introduce this change in a followup PR.

src/views/scans/__tests__/__snapshots__/showScansModal.test.tsx.snap
```diff
@@ -47,19 +47,60 @@ exports[`ShowScansModal should render a basic component: basic 1`] = `
...
   variant="medium"
 >
-  <Bullseye>
-    <EmptyState>
-      <EmptyStateHeader
-        headingLevel="h2"
-        icon={
-          <EmptyStateIcon
-            icon={[Function]}
+  <React.Fragment>
```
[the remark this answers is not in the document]
cdcabrera: This fallback is the cause @nicolearagao 

`(scanJobs || [])` + a `truthy` check for `sortedScanJobs &&` = `true` if you want to retain the `false` you'll now need to check for `length`

cdcabrera (approved):
The last commit resolves your self-find on the `.length` check. The work itself is lgtm!

PR #494, opened 2024-10-18
fix(viewSourcesList): ds-813 useShowConnectionsApi catch errors

cdcabrera:
## What's included

- fix(viewSourcesList): ds-813 useShowConnectionsApi catch errors

### Notes
- applies a catch similar to the `addSourceModal` implementation of `getCredentials`

### Local run check
1. `$ npm run start:stage`
1. you can attempt to emulate the behavior of what the display will look like on failure either
   - by mocking an error for the API call
   - or more directly by putting the same value used for the error inside the success callback, temporarily
   
   You should see a result similar to the example below on failure

## Example

Screenshot 2024-10-18 at 3 23 11 PM

## Updates issue/story

[DISCOVERY-813](https://issues.redhat.com/browse/DISCOVERY-813)

### DIFF
--- a/src/views/sources/viewSourcesList.tsx
+++ b/src/views/sources/viewSourcesList.tsx
@@ -303,7 +303,14 @@ const SourcesListView: React.FunctionComponent = () => {
       <Button
         variant={ButtonVariant.link}
         onClick={() => {
-          showConnections(source).then(success => setConnectionsData(success));
+          showConnections(source)
+            .then(success => setConnectionsData(success))
+            .catch(err => {
+              setConnectionsData(emptyConnectionData);
+              if (!helpers.TEST_MODE) {
+                console.error(err);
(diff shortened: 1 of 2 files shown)


cdcabrera:
After reviewing this, most of this update is unnecessary. The pr will be reworked to only broadcast the `console.error`

nicolearagao (approved):

nicolearagao:
> After reviewing this, most of this update is unnecessary. The pr will be reworked to only broadcast the `console.error`

@cdcabrera why ? I thought this solution was ok.

cdcabrera:
@nicolearagao it's unnecessary. the default state of `setConnectionsData` is already `emptyConnectionData`, having the `catch(err => { setConnectionsData(emptConnectionData` produces the same result as not having it

PR #496, opened 2024-10-22
fix(scans) ds-531 Stop showing download button too soon

mirekdlugosz:
Scan Download button is displayed when scan has report_id field. However, report_id is assigned as soon as scan finished running, but before fingerprinting phase has finished. And it is not possible to download a scan before fingerprinting phase finished.

Arguably this is API not playing very nice, but it's trivial for UI to wait for end_time field. That field is only present on scans that are all completed and reports can be downloaded.

Also, fix typing warning in the same file.

Relates to JIRA: DISCOVERY-531


cdcabrera (approved):
lgtm

PR #497, opened 2024-10-22
fix(sources): ds-436 display message if scan has not run yet

mirekdlugosz:
Fixes a minor issue from the doc:

> Sources [Mirek] Sources page may display empty cell in “Last connected” column.
> * Steps: Create a source. Click “Scan” to scan it. Wait for scan to finish and go to “Scans”. Find that scan and remove it. Go back to “Sources” page.

This also happens if you create a source using CLI or API directly, where you can opt-out of initial scan.

I'm happy to take suggestions how phrasing can be improved.

Not a huge fan of how this text is misaligned compared to text displayed when scan is running or finished (see screenshot). But for "scan-ful" sources text is displayed inside `<Button>`, which adds left padding, and as far as I can tell, we can't have "transparent" buttons (i.e. apply all the paddings, but none of text formatting). Given that most of users are never going to see that text, it's probably best to leave things like that.

Screenshot from 2024-10-22 15-33-36

Relates to JIRA: DISCOVERY-436

### DIFF
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -188,6 +188,7 @@
     "label_status_pending": "In progress",
     "label_status_running": "In progress",
     "label_status_paused": "Paused",
+    "label_status_not_started_sources": "Scan not started",
     "label_status_completed_scan": "Completed",
     "label_status_completed_sources": "Last connected",
     "label_status_failed_sources": "Connection failed",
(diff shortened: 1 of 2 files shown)


public/locales/en.json
```diff
@@ -188,6 +188,7 @@
     "label_status_pending": "In progress",
     "label_status_running": "In progress",
     "label_status_paused": "Paused",
+    "label_status_not_started_sources": "Scan not run yet",
```
cdcabrera, line 191: optional: run this by ccs/docs

cdcabrera (approved):
technically functional! lgtm! The copy might need to be iterated on

public/locales/en.json
mirekdlugosz: Docs suggested "Scan not started", which I now use in fd447c501c5a4852f581ee2e81b41e8adfb3485f

PR #498, opened 2024-10-23
feat(login): ds-805 Display proper message on 403 response

mirekdlugosz:
extractErrorMessage helper expects response body data, not generic AxiosError.

I tried, and failed, to check what kind of error message would be displayed if something happens to session while you are logged in. I'm happy to take pointers to other places similar change should be made.

Error message on bad credentials (this one comes from [locale file](https://github.com/quipucords/quipucords-ui/blob/main/public/locales/en.json#L135)):
Screenshot from 2024-10-23 11-11-45

Error message after providing bad credentials too many times (this one comes from backend):
Screenshot from 2024-10-23 11-11-54

Relates to JIRA: DISCOVERY-805

### DIFF
--- a/src/components/login/login.tsx
+++ b/src/components/login/login.tsx
@@ -62,7 +62,7 @@ const Login: React.FC<LoginProps> = ({ children, useGetSetAuth = useGetSetAuthAp
             error => {
               setLoginError(
                 (error?.status === 400 && t('login.error', { context: error?.status })) ||
-                  apiHelpers.extractErrorMessage(error)
+                  apiHelpers.extractErrorMessage(error?.response?.data)
               );
             }
           )


src/components/login/login.tsx
```diff
@@ -62,7 +62,7 @@ const Login: React.FC<LoginProps> = ({ children, useGetSetAuth = useGetSetAuthAp
             error => {
               setLoginError(
                 (error?.status === 400 && t('login.error', { context: error?.status })) ||
-                  apiHelpers.extractErrorMessage(error)
+                  apiHelpers.extractErrorMessage(error?.response?.data)
```
cdcabrera, line 65: optional:
**If a future goal is to always return hardcoded non-localized strings from the API** you could remove the example code used for the `400` and just do

```
setLoginError(apiHelpers.extractErrorMessage(error?.response?.data || t('login.error-fallback')))
```

**If a future goal is to use locale based in the frontend** then you can leave the example as a reference which was intended to highlight how you could offset error codes from the API (the http status was used because it was convenient). This solution would look something like the pseudo code

```
t('login.error', { context: [SOME CODE FROM THE API OR A LIST OF CODES] }))

//or

[LIST OF API CODES].map(CODE => t('login.error', { context: CODE }))
```

cdcabrera (approved):

src/components/login/login.tsx
mirekdlugosz: Judging by the vibe on a meeting today, generally we would prefer for backend to provide these messages. But that was not something we really discussed explicitly and at detail.

Ideally backend messages would be localized, although we don't have infrastructure for that at the moment and I think it's so low on priorities list it's not even in backlog.


> **If a future goal is to use locale based in the frontend** then you can leave the example as a reference which was intended to highlight how you could offset error codes from the API (the http status was used because it was convenient). This solution would look something like the pseudo code
> 
> ```
> t('login.error', { context: [SOME CODE FROM THE API OR A LIST OF CODES] }))
> 
> //or
> 
> [LIST OF API CODES].map(CODE => t('login.error', { context: CODE }))
> ```

I think we won't be able to do this exact thing because there might be various reasons for any given response code. Like 403 right now is only because user is locked out, but there are many possible reasons why server would decide to send back 403 code.

I am going to capture these thoughts in Jira and / or roadmap doc. For now team considers this to be a release blocker, so I'm going to merge current version and hopefully we can iterate over that at some point in the future.

PR #506, opened 2024-10-29
fix(sources): ds-436 accept wider range of directories values

mirekdlugosz:
I noticed that search directories textarea ("deep scan") has the same problem as hosts textarea. We fixed hosts in #475 (b1909a77b458f8be618dea023519fce1611a1230), now it's time to fix that one.

As far as I can tell, there are no other fields that would take comma-separated list of values.

I decided to rename `normalizeHosts` helper into something more generic and re-use that. This will conflict badly with #492 . I guess the alternative is to have internal helper `normalizeCommaSeparated` and public-facing `normalizeHosts` and `normalizeSearchDirs`. Not sure which solution is more aligned with current code practices.

### DIFF
--- a/src/helpers/__tests__/helpers.test.ts
+++ b/src/helpers/__tests__/helpers.test.ts
@@ -125,33 +125,33 @@ describe('formatDate', () => {
   });
 });
 
-describe('normalizeHosts', () => {
+describe('normalizeCommaSeparated', () => {
   it('should accept values separated by space', () => {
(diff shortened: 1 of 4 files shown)


cdcabrera (approved):
appears to work!

regarding the naming and conflict with PR #492 ... it's the developers preference, if it helps my vote is for getting the work done ... based on your choice of the generic naming that works! you can always rename or refactor later

PR #508, opened 2024-10-31
fix(sources): ds-436 wrap scan not run message in Button

mirekdlugosz:
This is followup to fd447c501c5a4852f581ee2e81b41e8adfb3485f (#497). After that commit, "Scan not started" message was not aligned with other statuses. Now message is wrapped in disabled Button and has an icon, making a look more consistent.

Screenshot from 2024-10-31 12-41-08

Relates to JIRA: DISCOVERY-436


nicolearagao (approved):
thanks mirek

PR #510, opened 2024-11-04
fix(views): ds-832 unselect items upon deletion

mirekdlugosz:
Selecting multiple items and deleting them kept them selected in state, which could result in failures of subsequent operations - especially creating scans (because scan would try to involve sources that no longer exist).

Since bulk delete might result in deleting some items, but not the others, arguably these that were not deleted should still be selected. With current code they won't. But that seems like a relatively minor inconvenience at best.

Relates to JIRA: DISCOVERY-832

### DIFF
--- a/src/views/credentials/viewCredentialsList.tsx
+++ b/src/views/credentials/viewCredentialsList.tsx
@@ -170,7 +170,7 @@ const CredentialsListView: React.FunctionComponent = () => {
   });
 
   const {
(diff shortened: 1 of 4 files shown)


infinitewarp (approved):
This seems reasonable to me! 😀  But definitely wait for a more UI-oriented expert to chime in before merging.

nicolearagao (approved):
lgtm

PR #512, opened 2024-11-06
chore: add a utility script to make development easier

bruno-fs:
`npm run start:using-server` will serve the development UI properly wired to a running instance of quipucords server. Default configuration points to quipucords-installer defaults, and can be customized with QUIPUCORDS_SERVER_* variables.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "build:pre": "bash ./scripts/pre.sh",
     "release": "changelog --link-url https://github.com/quipucords/quipucords-ui.git",
     "start": "export PROTOCOL=http; export HOST=127.0.0.1; export PORT=${PORT:-3000}; export MOCK_PORT=${MOCK_PORT:-3030}; run-p -l api:dev start:js start:open",
+    "start:using-server": "export PROTOCOL=${QUIPUCORDS_SERVER_PROTOCOL:-https}; export HOST=${QUIPUCORDS_SERVER_HOST:-127.0.0.1}; export PORT=${PORT:-3000}; export MOCK_PORT=${QUIPUCORDS_SERVER_PORT:-9443}; run-p -l start:js start:open",
     "start:js": "export NODE_ENV=development; weldable -l ts -x ./config/webpack.dev.js",
     "start:open": "xdg-open $PROTOCOL://$HOST:$PORT/ || open $PROTOCOL://$HOST:$PORT/",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -28,6 +28,7 @@
     "build:pre": "bash ./scripts/pre.sh",
     "release": "changelog --link-url https://github.com/quipucords/quipucords-ui.git",
     "start": "export PROTOCOL=http; export HOST=127.0.0.1; export PORT=${PORT:-3000}; export MOCK_PORT=${MOCK_PORT:-3030}; run-p -l api:dev start:js start:open",
+    "start:using-server": "export PROTOCOL=${QUIPUCORDS_SERVER_PROTOCOL:-https}; export HOST=${QUIPUCORDS_SERVER_HOST:-127.0.0.1}; export PORT=${PORT:-3000}; export MOCK_PORT=${QUIPUCORDS_SERVER_PORT:-9443}; run-p -l start:js start:open",
```
cdcabrera, line 31: some level of alteration is needed, a couple of choices for you...

**Option 1:** **highly recommend**
- combining the original and new script. reason, currently, the new script is customized for a single developer where as the original `start` script was meant to be expandable for multiple use cases, modifying the original would be natural
  
   technically the `npm-run-all` is just looking for `strings` to run `npm scripts` so replacing it with an exported empty command means it skips it... 
   ```
      "start": "export PROTOCOL=${DEV_SERVER_PROTOCOL:-http}; export HOST=${DEV_SERVER_HOST:-127.0.0.1}; export PORT=${DEV_SERVER_PORT:-3000}; export MOCK_PORT=${DEV_SERVER_MOCK_PORT:-3030}; export SERVER_SCRIPT=${DEV_SERVER_SCRIPT:-api:dev}; run-p -l $SERVER_SCRIPT start:js start:open",
   ```

**Option 2:** **if the unique script is kept...**
- `start:using-server` should be renamed...
   - reason, description is generic. technically the original script is also running various servers
   - recommend something like `start:custom-server` or something direct like `start:custom` since having to continuously type something with a hyphen gets tedious, for those who still reboot their computers =)

**And an additional recommendation, but not necessary**
- the prefix `QUIPUCORDS_SERVER_` should probably be shortened
   - reason, if the values using `QUIPUCORDS_SERVER_` are expected to be typed out by follow up developers we could be kind by shorten them. But, if there's another script exporting these values this is less of a concern.

cdcabrera (changes requested):
just a couple of choices, and a non-blocking recommendation

cdcabrera:
Another recommendation is to update the commit from `chore` to `build` since you're actively affecting the scripts that let us run and build everything

infinitewarp (approved):

nicolearagao (approved):
LGTM

PR #513, opened 2024-11-06
feat(ActionMenu<Scan>): ds-437 add download report button

bruno-fs:
The button shall only be enabled when the scan is completed.

## What's included

Includes a dropdown button on scan row kebab for downloading reports.

### Notes
- The button should be disabled for failed or incomplete scans.
- If a "complete" scan doesn't have a `report_id` (due to a bug/edge-case on the backend?). a client-side error will be thrown

## Example
Kebab for scans ~in progress~ not ready for download
image
...and for completed ones
image
On download completion, an alert should be displayed. Same shall happen on errors, but that's harder to test without
monkey-patching the server
image

...

## Updates issue/story

[DISCOVERY-347](https://issues.redhat.com/browse/DISCOVERY-347)

### DIFF
--- a/src/components/actionMenu/actionMenu.tsx
+++ b/src/components/actionMenu/actionMenu.tsx
@@ -11,7 +11,7 @@ import { EllipsisVIcon } from '@patternfly/react-icons';
 
 interface ActionMenuProps<T = unknown> {
   item: T;
-  actions: { label: string; onClick: (item: T) => void }[];
+  actions: { label: string; onClick: (item: T) => void; disabled?: boolean }[];
 }
 
 const ActionMenu = <T,>({ item, actions }: ActionMenuProps<T>) => {
@@ -43,6 +43,7 @@ const ActionMenu = <T,>({ item, actions }: ActionMenuProps<T>) => {
             onClick={() => {
               a.onClick(item);
             }}
+            isDisabled={a.disabled}
           >
             {a.label}
           </DropdownItem>
--- a/src/helpers/helpers.ts
+++ b/src/helpers/helpers.ts
@@ -9,7 +9,7 @@ import React from 'react';
 import moment, { type MomentInput } from 'moment';
 import titleImg from '../images/title.svg';
 import titleImgBrand from '../images/titleBrand.svg';
-import { type CredentialType } from '../types/types';
+import { type CredentialType, type MostRecentScan, type scanJob } from '../types/types';
 
 /**
  * Is dev mode active.
@@ -226,8 +226,22 @@ const getCurrentDate = () => (TEST_MODE && moment.utc('20241001').toDate()) || m
  */
 const getTitleImg = (isBrand = UI_BRAND) => ((isBrand && titleImgBrand) || titleImg) as string;
 
+/**
+ * Return if a report associated with given ScanJob can be downloaded
+ *
+ * @param {scanJob | MostRecentScan} job
+ * @returns {boolean}
+ */
+const canDownloadReport = (job?: scanJob | MostRecentScan) => {
+  if (job && job.status === 'completed' && job.report_id) {
+    return true;
+  }
+  return false;
+};
+
 const helpers = {
   authType,
+  canDownloadReport,
   downloadData,
   noopTranslate,
   generateId,
--- a/src/helpers/helpers.ts
+++ b/src/helpers/helpers.ts
@@ -232,12 +232,8 @@ const getTitleImg = (isBrand = UI_BRAND) => ((isBrand && titleImgBrand) || title
  * @param {scanJob | MostRecentScan} job
  * @returns {boolean}
  */
-const canDownloadReport = (job?: scanJob | MostRecentScan) => {
-  if (job && job.status === 'completed' && job.report_id) {
-    return true;
-  }
-  return false;
-};
+const canDownloadReport = (job?: scanJob | MostRecentScan) =>
+  job?.status === 'completed' && job?.report_id !== undefined;
 
 const helpers = {
   authType,
--- a/src/types/types.ts
+++ b/src/types/types.ts
@@ -174,12 +174,16 @@ export type ScanOptions = {
 export type scanJob = {
   id: number;
   report_id: number;
+  status: string;
 };
 
 export type StatusDetails = {
   job_status_message: string;
 };
 
+// TODO: the object returned from the API representing scanJob and MostRecentScan are the same;
+// consider this in a future refactor. Also, MostRecentScan is a ScanJob, not a scan. Yeah, quipucords
+// is confusing...
 export type MostRecentScan = {
   id: number;
   report_id: number;
--- a/src/types/types.ts
+++ b/src/types/types.ts
@@ -181,9 +181,9 @@ export type StatusDetails = {
   job_status_message: string;
 };
 
-// TODO: the object returned from the API representing scanJob and MostRecentScan are the same;
-// consider this in a future refactor. Also, MostRecentScan is a ScanJob, not a scan. Yeah, quipucords
-// is confusing...
+// TODO: Considerations for a future refactor: merge and/or rename scanJob and MostRecentScan.
+// Reason: the object returned from the api representing scanJob and MostRecentScan are the same;
+// also, in quipucords jargon, MostRecentScan is a ScanJob, not a Scan.
 export type MostRecentScan = {
   id: number;
   report_id: number;
--- a/src/views/scans/viewScansList.tsx
+++ b/src/views/scans/viewScansList.tsx
@@ -250,6 +250,15 @@ const ScansListView: React.FunctionComponent = () => {
                             setScanSelected(undefined);
                           });
                         }
+                      },
+                      {
+                        label: t('table.label', { context: 'download' }),
+                        disabled: !helpers.canDownloadReport(scan?.most_recent),
+                        onClick: () => {
+                          if (scan?.most_recent) {
+                            downloadReport(scan.most_recent.report_id);
+                          }
+                        }
                       }
                     ]}
                   />
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -173,6 +173,7 @@
     "label_satellite": "Satellite",
     "label_rhacs": "RHACS",
     "label_rescan": "Rescan",
+    "label_download": "Download",
     "label_scan": "Scan",
     "label_source": "Source",
     "label_source_other": "Sources",
--- a/src/views/scans/__tests__/showScansModal.test.tsx
+++ b/src/views/scans/__tests__/showScansModal.test.tsx
@@ -16,7 +16,7 @@ describe('ShowScansModal', () => {
         scanJobs={[
           {
             id: 12345,
-            status: 'DOLOR SIT',
+            status: 'completed',
             start_time: new Date('2024-09-06'),
             end_time: new Date('2024-09-06'),
             report_id: 67890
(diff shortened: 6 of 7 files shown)


gtanzillo:
Thanks for adding this @bruno-fs. Let try to get this one in for the v2 release, since it addresses a blocker issue.

bruno-fs:
After sleeping on it, I believe we could test this workflow directly in the camayoc, given that we were missing the step to complete the standard workflow for scanning.

Thoughts, @mirekdlugosz & @ruda ?

cdcabrera (changes requested):
nice start! you provided your own direct solution...

src/hooks/useScanApi.ts
```diff
@@ -343,8 +343,19 @@ const useDownloadReportApi = (onAddAlert: (alert: AlertProps) => void) => {
...
+    });
+    return;
+  }, [onAddAlert, t]);
+
   const downloadReport = useCallback(
-    async (reportId: number) => {
+    async (reportId: number | undefined) => {
+      if (!reportId) {
+        return callbackErrorNoReport();
+      }
```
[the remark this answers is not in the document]
bruno-fs: Good idea! thanks for the feedback!

src/hooks/useScanApi.ts
```diff
@@ -343,8 +343,19 @@ const useDownloadReportApi = (onAddAlert: (alert: AlertProps) => void) => {
...
+    onAddAlert({
+      title: t('toast-notifications.description_report_downloaded_error_no_report'),
+      variant: 'danger'
+    });
+    return;
+  }, [onAddAlert, t]);
+
   const downloadReport = useCallback(
-    async (reportId: number) => {
+    async (reportId: number | undefined) => {
```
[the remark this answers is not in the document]
bruno-fs: Taking a tangent here, I wrote a function that might suffer from this problem too

```typescript
const canDownloadReport = (job: scanJob | MostRecentScan) => {
  if (job.status === 'completed' && job.report_id) {
    return true;
  }
  return false;
};
```
I considered adding `undefined` to the signature so I wouldn't need to check type before invoking the function, but I remembered this comment.

I'd personally to have the function checking if `undefined` (and returning `false` immediately) then writing 

`job && canDownloadReport(job)` or 

```
if (job) { canDownloadReport(job }
```

Do you have any advice here?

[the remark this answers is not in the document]
bruno-fs: mmm for now I'll stick with

```
const canDownloadReport = (job?: scanJob | MostRecentScan) => {
  if (job && job.status === 'completed' && job.report_id) {
    return true;
  }
  return false;
};
```

src/hooks/useScanApi.ts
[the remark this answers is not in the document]
bruno-fs: replacing `isDisabled` with `disabled` is proving to be tricky. If I replace `isDisabled` with `disabled` it doesn't work 🤔 

Any idea why?

src/components/actionMenu/actionMenu.tsx
```diff
@@ -43,6 +43,7 @@ const ActionMenu = <T,>({ item, actions }: ActionMenuProps<T>) => {
             onClick={() => {
               a.onClick(item);
             }}
+            isDisabled={a.disabled}
```
bruno-fs, line 46: Here. If I change this to `disabled` I end up with download buttons always enabled :/

src/hooks/useScanApi.ts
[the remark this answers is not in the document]
bruno-fs: https://github.com/quipucords/quipucords-ui/pull/513/files#r1834930552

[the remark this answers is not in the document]
cdcabrera: apologies for the confusion @bruno-fs ... `isDisabled` was a reference to what `disabled` was doing behind the scenes... I'll take another look since you appear to have updated/added something, but changing your original "disabled" to `disabled: scan.most_recent?.report_id !== undefined && scan.most_recent?.status !== 'completed',` would have produced the results you wanted

src/helpers/helpers.ts
```diff
@@ -226,8 +226,22 @@ const getCurrentDate = () => (TEST_MODE && moment.utc('20241001').toDate()) || m
...
+ *
+ * @param {scanJob | MostRecentScan} job
+ * @returns {boolean}
+ */
+const canDownloadReport = (job?: scanJob | MostRecentScan) => {
+  if (job && job.status === 'completed' && job.report_id) {
+    return true;
+  }
+  return false;
+};
```
cdcabrera, line 240: required: simplify, no need to return a `boolean` when the conditions themselves can return a `boolean`

change to either
```
const canDownloadReport = (job?: scanJob | MostRecentScan) =>
  job?.status === 'completed' && typeof job?.report_id === 'number';
```
or
```
const canDownloadReport = (job?: scanJob | MostRecentScan) =>
  job?.status === 'completed' && job?.report_id !== undefined;
```

src/hooks/useScanApi.ts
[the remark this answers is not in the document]
bruno-fs: Alright, no problem! 

I just added an extra commit that should hopefully pass all local testing - it will fail commit linting because it is a fixup! commit that I'll squash later if everything is OK.

src/views/scans/viewScansList.tsx
```diff
@@ -255,7 +255,9 @@ const ScansListView: React.FunctionComponent = () => {
                         label: t('table.label', { context: 'download' }),
                         disabled: !helpers.canDownloadReport(scan?.most_recent),
                         onClick: () => {
-                          downloadReport(scan.most_recent.report_id);
+                          if (scan?.most_recent) {
```
bruno-fs, line 258: I had to add this condition so typescript stop complaining, but it seems pointless given the button will be disabled if `scan.most_recent` is `undefined`. 

Is there a better way to handle these situations, @cdcabrera?

src/components/actionMenu/actionMenu.tsx
cdcabrera: i'm not sure what that means, but this is part of the solution, leave as is

src/types/types.ts
```diff
@@ -174,12 +174,16 @@ export type ScanOptions = {
...
+  status: string;
 };
 
 export type StatusDetails = {
   job_status_message: string;
 };
 
+// TODO: the object returned from the API representing scanJob and MostRecentScan are the same;
+// consider this in a future refactor. Also, MostRecentScan is a ScanJob, not a scan. Yeah, quipucords
+// is confusing...
```
cdcabrera, line 186: optional: may reconsider and remove the personal touch on the TODO

cdcabrera (changes requested):
despite the confusion almost there, minor changes to the helper

src/types/types.ts
cdcabrera: sure

...what you had originally worked too... you could have just removed the last sentence

src/views/scans/viewScansList.tsx
bruno-fs: cool. thanks for the alternatives!

mirekdlugosz:
> After sleeping on it, I believe we could test this workflow directly in the camayoc, given that we were missing the step to complete the standard workflow for scanning.

Missed that comment.

First, so we are on the same page: Camayoc covers scanning and downloading scan, in both old and new UI. In new UI it will click button in "Most recent" column and try to download scan from there. That was the only way to get the report before this PR.

With that PR we get two paths to download reports, and I'm not sure if we should cover both in Camayoc, or only one - and which one. I think I would like to better understand what we can cover in unit tests here before making a final decision. That's probably something we will have to decide on post new UI release.

mirekdlugosz:
I know code is still subject to change, but I verified that locally and it works as intended. "Download" is disabled for scans that are not completed (=never run, running right now, canceled, failed). If there are multiple scan jobs, it downloads the latest one. The same report is downloaded when using kebab menu item and button inside modal.

src/helpers/helpers.ts
cdcabrera: With the update... I'm not sure this is what we wanted, the return value on this would helper would actually be `boolean | number | undefined`... adding in all of what was recommended is preferable, either the `typeof === number` or `!== undefined` evaluation... 

...without adding in the additional evaluation this inline condition will return whatever the `report_id` value is instead of an actual `boolean`. If the first value is `true` (ie. `boolean && number|undefined`...) the boolean passes the ternary AND will return the last value, which could be `truthy` or `falsey`.

JS can be deceptive. I recommend you run each of these lines in the browser console dev tools
```
// this should return 1234
true && 1234 
// this should return false
false && 1234
// this should return undefined
true && undefined 
// this should return false since 
false && undefined
```

mirekdlugosz:
This is tangentially related, but I believe this change reveals a problem that we had for a long time.

When clicking the kebab icon second time, dropdown menu is positioned slightly off the viewport, causing entire table to shift a bit to the left. Easy to reproduce in `npm run start` - just go to "Scans" and keep clicking any kebab icon. Other views are not affected - I guess it has to do with length of strings inside dropdown.

Opened [DISCOVERY-846](https://issues.redhat.com/browse/DISCOVERY-846) with a recording and more details.

I wonder if this is easy enough to fix that we could introduce it as part of this PR? If not, I guess this will be on a growing list of issues to fix sometime after initial new UI release...

bruno-fs:
@mirekdlugosz 

>I wonder if this is easy enough to fix that we could introduce it as part of this PR? If not, I guess this will be on a growing list of issues to fix sometime after initial new UI release...

Might be easy, but I've no idea how to fix this 😅 
Is it a regression? Considering I'll be out for a few days I guess someone else will need to fix it.

src/helpers/helpers.ts
bruno-fs: oh, sorry. I thought I added exactly what you recommended. changing it now.

cdcabrera: With the update, very close. Looks like we missed another linting setting`'arrow-body-style': 2,`

To align with our other "arrow functions", per the 2 examples, you don't have to include the `return`... removing it aligns to the style choice we've made. 
```
const canDownloadReport = (job?: scanJob | MostRecentScan) =>
  job?.status === 'completed' && job?.report_id !== undefined;
```

We'll add the `'arrow-body-style': 2,` into the linting rules in a subsequent PR since running the linter adds files that are outside of this PRs work

cdcabrera (changes requested):
align the arrow func helper style ~remove the unneeded test update~

src/helpers/helpers.ts
bruno-fs: oh shoot 😅

do you mind changing this commit or asking someone on the team to do it? I'm PTO now 😎

src/hooks/useScanApi.ts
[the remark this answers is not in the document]
mirekdlugosz: There was a lot of traffic in this PR and I missed that topic. Sorry!

I think it's fine if `canDownloadReport` allows undefined as an input. After all, the entire point of the function is to guard `downloadReport`. So as long as `canDownloadReport` returns false in all cases where report can't be downloaded (including various objects being undefined), I would say we are good.

src/helpers/helpers.ts
mirekdlugosz: I applied this change in 2ea72786d08a9848e67d91114bf08f0cf67834e6 . I have not touched linting rules, though I see so far we don't have these included in any other PR.

mirekdlugosz:
@cdcabrera from your point of view, is there anything else to be done here?

I applied last remaining comment and pushed changes. Now all conversations are resolved. I also did one more round of testing of actual build, and did not see anything that would require a fix before release.

cdcabrera (approved):
styling aligns now! lgtm

nicolearagao (approved):

PR #511, opened 2024-11-06
feat(addCredentialModal): don't allow submission of form with errors

bruno-fs:
This is a PoC on how to solve DISCOVERY-839. Tests weren't updated to reflect changes yet - I'd like an early feedback from our experts @nicolearagao and @cdcabrera before proceeding

## What's included
- Don't allow submission of credential form if required data is missing
- Additionally, render errors captured from backend validation errors inline and don't allow submitting the form if the offending value isn't modified.

### Notes

"Save" button is disabled until the required data is properly set
image
Filling required data enables the "Save" button
image
If a required field is erased, an inline error is displayed and "Save" button is disabled again
image

If the backend responds with validation errors, those should be rendered inline, and the submission button shall be disabled until the value is modified
image

## Example

...

## Updates issue/story

[DISCOVERY-839](https://issues.redhat.com/browse/DISCOVERY-839)

Relates to JIRA: DISCOVERY-839

PR #517, opened 2024-11-14
fix: ds-839 block modal form submissions

cdcabrera:
## What's included

- fix: ds-839 block modal form submissions

### Notes
- leverages existing toast notification layout to aid in preventing modal closing after form submission
- basic blocking for modal forms for
   - Add/Edit source api hooks
   - Add scan api hooks
   - Add/Edit credential api hooks
      - Technically the Edit aspect is still being worked on under #428  this simply adds the `Promise.reject` when ready
- what this doesn't do
   - enhance the toast notification messages or perform inline form messaging.
   - give some level of loading indicator for the form submission

### Local run check
1. `$ npm run start:stage`
1. confirm basic form submission for 
   - Add/Edit sources is blocked with a toast notification
   - Adding a scan is blocked with a toast notification
   - Adding a credential is blocked with a toast notification

## Example

expected basic toast notifications
Screenshot 2024-11-14 at 2 06 55 PM (2)
Screenshot 2024-11-14 at 2 06 39 PM (2)
Screenshot 2024-11-14 at 2 06 12 PM (2)

## Updates issue/story

[DISCOVERY-839](https://issues.redhat.com/browse/DISCOVERY-839)

### DIFF
--- a/src/hooks/__tests__/__snapshots__/useCredentialApi.test.ts.snap
+++ b/src/hooks/__tests__/__snapshots__/useCredentialApi.test.ts.snap
@@ -11,7 +11,7 @@ exports[`useAddCredentialApi should attempt an api call to add credentials: apiC
 ]
 `;
 
-exports[`useAddCredentialApi should handle errors while attempting to add a credential: addCredentials, error 1`] = `
+exports[`useAddCredentialApi should handle errors while attempting to add a credential: addCredentials, alert error 1`] = `
 [
   [
     {
@@ -27,6 +27,13 @@ exports[`useAddCredentialApi should handle errors while attempting to add a cred
     },
   ],
 ]
+`;
+
+exports[`useAddCredentialApi should handle errors while attempting to add a credential: addCredentials, error 1`] = `
+{
+  "isAxiosError": true,
+  "message": "Mock error",
+}
 `;
 
 exports[`useAddCredentialApi should handle success while attempting to add a credential: addCredentials, success 1`] = `
@@ -46,6 +53,16 @@ exports[`useAddCredentialApi should handle success while attempting to add a cre
 `;
 
 exports[`useAddCredentialApi should process an API error response: callbackError 1`] = `
+{
+  "response": {
+    "data": {
+      "message": "Dolor sit",
+    },
+  },
+}
+`;
+
+exports[`useAddCredentialApi should process an API error response: callbackError, alert 1`] = `
 [
   [
     {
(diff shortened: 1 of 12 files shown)


mirekdlugosz (changes requested):
I have tested that locally. Covered create credential modal, create source modal, edit source modal and new scan modal. Edit credential is worked on in another PR, we will need to test that modal once at least one of PRs is merged and other is rebased.

Overall it works as we intended for this stage. There are three problems I have noticed, only first of which I would consider required. I'm putting all three here in case they can be fixed cheaply while doing the required one. If not, let me know, I will create tickets for them and we will handle them in due time.

### Problem 1 (required)

If I don't fill scan name, there are going to be **two** toast notifications.

Steps:

1. Create credential and scan of any type
2. Go to Sources and click "Scan" link for any source
3. Leave "Name" field empty
4. Click Save

Result:

Screenshot from 2024-11-18 19-11-08

### Problem 2 (optional)

Error messages from server are processed before being displayed. As a result there are HTML entity codes and backslash character may be missing.

Steps:

1. Create credential of network type
2. Go to Sources and click "Add Source" -> "Network range"
3. Fill all fields
4. In "Search addresses", insert `192.1.5.1/`

Result:
(note `#x2F;` and that there is single backslash instead of two)

Screenshot from 2024-11-18 18-34-13

### Problem 3 (optional)

Field names from server errors are missing. Error messages are constructed by [`extractErrorMessage`](https://github.com/quipucords/quipucords-ui/blob/main/src/helpers/apiHelpers.ts), but that one looks only into response values - while response object keys contain important contextual information.

At one point we have discussed introducing a copy of `extractErrorMessage` function, but one that would also display error keys. The idea would be that we don't want to modify existing function due to regression testing surface that would require. But not sure if it's something we can do, as I don't see `extractErrorMessage` called explicitly in this PR anywhere.

Steps:

1. Go to Credentials and click "Add Credential" -> "Network range"
2. Leave form empty and click "Save"

Result (no duplicated "This field is required.")

Screenshot from 2024-11-18 19-10-18

cdcabrera:
@mirekdlugosz not going to be expanding this PR

reasons
- Technically this work still has to happen regardless of how advanced or basic the end-result work is
- This PR is simply exposing what was already happening inside the UI

Iteration to maintain timeline is the key here

mirekdlugosz (approved):
@justinkreft and @gtanzillo agree this is not a release blocker. I have created a Jira [DISCOVERY-859](https://issues.redhat.com/browse/DISCOVERY-859) so we can fix that later.

For problem number 2, I have created Jira [DISCOVERY-856](https://issues.redhat.com/browse/DISCOVERY-856).

For problem number 3, I had hoped this is something we can improve cheaply before the release. Since we can't, I feel that after the release our efforts are better spent on putting these error messages near fields instead of toasts. Something like a work done in #511 . So there is no separate ticket for this.

PR #518, opened 2024-11-14
Test DISCOVERY-437 (bonus: smoke test viewScanList!)

bruno-fs:
## What's included
Add a sanity check testing viewScanList and the behavior of the kebab
button;
    
Properly test DISCOVERY-437; given how tighly coupled everything is,
adding a view test was the only way to test it.

### Notes
changing jest config and adding babel as transformer was required to be able to test those views - setting up those configs was way more challenging than actually writing those tests 😅 

## How to test
npm test

1. `$ npm test`
-->

## Example

...

## Updates issue/story


cdcabrera:
@bruno-fs the primary views were disabled from testing for multiple internal reasons discussed with @nicolearagao around how to architect UI code.

Any future "view" code actively needs to be broken apart into subsequent components, helpers, etc... before attempting to shoe-horn a testing solution into place. Some of the "many" future breakouts that need to happen include
- Toast notification alerts
- Table/inventory
- Moving API calls into the actual modals

...performing these types of updates will subsequently make testing the views easier. I recommend closing this PR down in the short term

cdcabrera (changes requested):
This work should not be merged until an overarching understanding of future UI architecture is discussed or understood

bruno-fs:
That's an interesting point, @cdcabrera. While I agree the views (or at least this one) desperately needs a refactor that makes it more decoupled - and at the very minimum easier to test, I don't think testing them (at least smoke testing) should be postponed.

Waiting to test later is a liability and should be a team decision, not up to just you and @nicolearagao.

bruno-fs:
@mirekdlugosz it seems it won't be possible to test download functionality from scans page on this codebase.

I wrote this one because of your previous comment about having download functionality already covered on camayoc, but on the scan jobs modal... Do you have a suggestion on how to test this while the refactor doesn't happen?

src/views/scans/__tests__/viewScansList.test.tsx
```diff
@@ -0,0 +1,166 @@
...
+    // repeat the process for other scans, which should not be downloadable
+    const otherScans = [1, 2, 3, 4];
+    otherScans.forEach(n => {
+      const { downloadBtn: button } = clickKebabMenu(`not downloadable ${n}`);
+      expect(button.hasAttribute('disabled')).toBeTruthy();
+      act(() => {
+        fireEvent.click(downloadBtn);
+      });
+      expect(mockedDownloadReport).not.toHaveBeenCalled();
+    });
```
bruno-fs, line 164: @cdcabrera just to be sure we are on the same page, if I remove all the proposed tests but this one (this was the true goal of testing here), would you still be against this PR?

mirekdlugosz:
@bruno-fs just FYI, covering both paths to download report (kebab menu and modal) is introduced in https://github.com/quipucords/camayoc/pull/557

PR #552, opened 2025-01-31
konflux: fix broken EC checks

bruno-fs:
A bug caused an update from konflux to remove this info.

### DIFF
--- a/.tekton/discovery-ui-pull-request.yaml
+++ b/.tekton/discovery-ui-pull-request.yaml
@@ -164,7 +164,7 @@ spec:
         - name: name
           value: init
         - name: bundle
-          value: quay.io/konflux-ci/tekton-catalog/task-init@sha256:4c6712db9419461b8c8a39523c012cb0dc061fb58563bb9170b3777d74f54659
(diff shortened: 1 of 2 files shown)


mirekdlugosz (approved):
Out of curiosity, does it matter? I would imagine if you provide image hash, then that is used to identify image to pull. So tag is only informative. And if you have multiple semantically different tags mapping to the same image hash, you probably have much larger problems.

In podman, if I try to pull image identified by both tag and hash, then `podman images` claims that image is not tagged. So it kind of looks like tag is ignored somewhere along the way. No idea if it is ignored in the part of stack that podman shares with OpenShift, or somewhere else.

Of course none of that prevents PR from being merged.

bruno-fs:
> Out of curiosity, does it matter? I would imagine if you provide image hash, then that is used to identify image to pull.

You are correct, @mirekdlugosz. The hash alone is enough for the images to be pulled and used on Konflux. However, the version does matter. Even though we can build everything successfully without those tags, later an EC check will fail because Red Hat EC expects them (kinda pointless if you ask me, the image hash should be enough).

PR #559, opened 2025-02-04
ci(konflux): dynamically set version labels

bruno-fs:
## What's included
~~Add a make target to bump quipucords-ui version~~
Dynamically set "version" (X.Y.Z) and "version_minor" (X.Y) labels based on the version set on `package.json`.

The labels added on Containerfile match the tags proposed for use on the `ReleasePlanAdmission` .
See https://url.corp.redhat.com/473b3be.

### DIFF
--- a/.tekton/discovery-ui-pull-request.yaml
+++ b/.tekton/discovery-ui-pull-request.yaml
@@ -6,6 +6,8 @@ metadata:
     build.appstudio.redhat.com/commit_sha: '{{revision}}'
     build.appstudio.redhat.com/pull_request_number: '{{pull_request_number}}'
     build.appstudio.redhat.com/target_branch: '{{target_branch}}'
+    # custom tasks defined in this repo https://pipelinesascode.com/docs/guide/resolver/#tasks-or-pipelines-inside-the-repository
+    pipelinesascode.tekton.dev/task: '[.tekton/task/generate-version-labels.yaml]'
     pipelinesascode.tekton.dev/max-keep-runs: "3"
     pipelinesascode.tekton.dev/on-cel-expression: event == "pull_request" && target_branch
       == "main"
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -5,10 +5,20 @@ help:
...
 
+bump-version:
+ifeq ($(version),)
+$(error Variable "version" is not set)
+endif
+	$(eval MINOR=$(shell echo $(version) | sed 's/\([0-9]*\.[0-9]*\).*$$/\1/'))
+	@echo "Bumping version to $(version)"
+	sed -i 's/^\(LABEL version=\).*$$/\1$(version)/' Containerfile
+	sed -i 's/^\(LABEL version_minor=\).*$$/\1$(MINOR)/' Containerfile
+	npm run release -- --override $(version)
```
infinitewarp, line 24: I believe this command generates a commit, but something else still needs to actually push it to the remote. Is anything pushing here? Or do you know if that's irrelevant for this process?

bruno-fs: oh, does npm release generate a commit? didn't knew that.

Oh crap, it needs to be ajusted then to git add Containerfile.

Answering your question, it is not pushing anything.

bruno-fs: maybe I should update the release script instead of doing this here 🤔

bruno-fs: But I'd still add as a Makefile target because my intention is to have this process streamlined for all repos

bruno-fs: mmm this is the changelog script, an external dep
https://github.com/cdcabrera/changelog-light

bruno-fs: I'm not seeing a way to add more files to the commit it will create. I think we will need to run it with `--commit=false` and start commiting the changes manually

bruno-fs: I completely altered the approach here, so I'm going to mark this discussion as resolved.
Feel free to reopen if you I missed anything, @infinitewarp

infinitewarp (approved):

PR #577, opened 2025-02-26
build: move to node 22

bruno-fs:
Node 18 support shall end in the next 2 months. We need to migrate.
We could move to version 20, but considering 22 is also a LTS version and its EOL is set to 2027, let's try that.

## Updates issue/story
[DISCOVERY-901](https://issues.redhat.com/browse/DISCOVERY-901)

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -14,7 +14,7 @@ jobs:
       contents: write
     strategy:
       matrix:
-        node-version: [18.x, 20.x]
+        node-version: [22.x]
     steps:
       - uses: actions/checkout@v4
       - name: Setup Node.js ${{ matrix.node-version }}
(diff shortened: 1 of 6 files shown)


infinitewarp (approved):

infinitewarp (changes requested):
Oh, I just remembered: Don't forget to update the README or other docs that direct people to use older Node versions.

like
> Your system needs to be running [NodeJS version 18+ and NPM](https://nodejs.org/)

bruno-fs:
Updated the README and package.json and its lock (with `npm install --package-lock-only`). Other references I found only refer to nodejs LTS https://github.com/quipucords/quipucords-ui/blob/3825c7c415ad8eb68d6604c0029e26d416ab0b3f/CONTRIBUTING.md#L258-L261

infinitewarp (approved):

PR #594, opened 2025-03-12
Backport make lock-baseimages to quipucords-ui

bruno-fs:
Add a make target for locking base images on demand.


infinitewarp (approved):
Locally running `podman build` and `make lock-baseimages` both succeeded without errors. 👍

abellotti (approved):
also successful gmake lock-baseimages on arm macbook.

PR #596, opened 2025-03-13
perf(lock-baseimages): speed up make lock-baseimages

bruno-fs:
Turns out pulling the image is an unecessary step. Skopeo can pull the metadata directly from the container repository.

Also run `make lock-baseimages` to keep the Containerfile fresh.


abellotti (approved):

abellotti:
successfully ran on macos 👍

PR #603, opened 2025-03-18
fix(scans): dsc-891 Use "scan" in scan removal dialog

mirekdlugosz:
Scan removal confirmation dialog should refer to "scan", not "source".

There was already a string for dialog body, but it was unused. That string contained `<0>` `</0>`, which I assume should be HTML elements to wrap scan name - but they were printed verbatim by UI. Also, neither credential nor source wraps their name in element, so there's consistency in not using it in scan.

Relates to JIRA: DISCOVERY-891

### DIFF
--- a/public/locales/en.json
+++ b/public/locales/en.json
@@ -8,13 +8,14 @@
     "username": "Username"
   },
   "form-dialog": {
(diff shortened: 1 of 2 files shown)


nicolearagao (approved):
LGTM

PR #605, opened 2025-03-24
feat(about): dsc-857 show CPU architecture

mirekdlugosz:
Add CPU Architecture field to "About" dialog. Data is already provided by a server. I decided to use [`platform.machine()`](https://docs.python.org/3/library/platform.html#platform.machine), the alternative being [`platform.processor()`](https://docs.python.org/3/library/platform.html#platform.processor). On upstream container both provide the same value, while on my local Python running on bare metal `processor()` is empty string.

I guess technically backend container might be using different arch than frontend container, but we don't support setups like that.

Relates to JIRA: DISCOVERY-857

Screenshot From 2025-03-24 14-31-41

### DIFF
--- a/src/components/aboutModal/aboutModal.tsx
+++ b/src/components/aboutModal/aboutModal.tsx
@@ -93,6 +93,16 @@ const AboutModal: React.FC<AboutModalProps> = ({
               </TextListItem>
             </React.Fragment>
           )}
+          {stats?.platform.machine && (
+            <React.Fragment>
+              <TextListItem className={loadingClassName} component="dt">
+                {t('about.server-cpu-arch')}
+              </TextListItem>
+              <TextListItem className={loadingClassName} component="dd">
+                {stats.platform.machine}
+              </TextListItem>
+            </React.Fragment>
+          )}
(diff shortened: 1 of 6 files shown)


src/components/aboutModal/aboutModal.tsx
```diff
@@ -93,6 +93,16 @@ const AboutModal: React.FC<AboutModalProps> = ({
               </TextListItem>
             </React.Fragment>
           )}
+          {stats?.platform.machine && (
```
mirekdlugosz, line 96: Technically, this may be an empty string. Do I need to add another condition to prevent it from showing, or does TypeScript automatically do the right thing here?

CC @nicolearagao

infinitewarp (approved):

src/components/aboutModal/aboutModal.tsx
nicolearagao: Empty string is considered falsy, the expression will short-circuit and nothing will be rendered. U are good!

nicolearagao (approved):

PR #630, opened 2025-05-19
feat(showAggregateReportModal): dsc-762 aggregate report summary

cdcabrera-test:
## What's included
- feat(showAggregateReportModal): dsc-762 aggregate report summary
- refactor(helpers): dsc-762 rename canDownloadReport
- feat(useScanApi): dsc-762 aggregate report api hook

## Notes
- Modal and view integration, align with existing action menu and modals
   - The associated "date-like-string" was passed through an existing display helper, `formatDate`
- Minor rename of helper to be more neutral/descriptive of use, `canDownloadReport` -> `canAccessMostRecentReport `
- API hook, align with existing api calls

### Pending
Screenshot 2025-05-19 at 3 11 11 PM

### Completed and modal display
Screenshot 2025-05-19 at 3 03 47 PM
May-19-2025 15-22-24

### 404
Screenshot 2025-05-19 at 3 05 00 PM

### 401
Redirects to login as intended

## Updates issue/story
[DISCOVERY-762](https://issues.redhat.com/browse/DISCOVERY-762)


bruno-fs (approved):
lgtm :shipit:

abellotti (approved):

cmchase:
Thanks @cdcabrera for the great description and extensive screenshots!

src/views/scans/showAggregateReportModal.tsx
```diff
@@ -0,0 +1,120 @@
...
+/**
+ * Apply list format, sort, and filter to report
+ */
+const formatSortFilterReportStats = (
+  report: PartialReportsAggregateResponse = {},
+  { filter = statsFilter }: { filter?: string[] } = {}
+) => {
+  const { results: resultStats = {}, diagnostics: diagnosticStats = {} } = report;
+
+  return [...(resultStats && Object.entries(resultStats)), ...(diagnosticStats && Object.entries(diagnosticStats))]
```
nicolearagao: Not a blocker, but have you considered keeping `results` and `diagnostics` separated in the UI instead of merging and sorting them together? Just wondering if there would be any downside or extra complexity in doing so.

src/views/scans/__tests__/showAggregateReportModal.test.tsx
```diff
@@ -90,4 +90,9 @@ describe('formatSortFilterReportStats', () => {
   it('should handle a missing report', () => {
     expect(formatSortFilterReportStats()).toMatchSnapshot('missing');
   });
+
+  it('should handle misuse', () => {
+    // @ts-expect-error Incorrect type, misuse or unknown passed to function
+    expect(formatSortFilterReportStats({ results: null, diagnostics: null })).toMatchSnapshot('misuse');
```
nicolearagao, line 96: Is the goal here to make sure the function can handle bad input like null values, or should this kind of check be handled before calling it?

src/views/scans/showAggregateReportModal.tsx
```diff
@@ -59,7 +59,7 @@ const formatSortFilterReportStats = (
 ) => {
   const { results: resultStats = {}, diagnostics: diagnosticStats = {} } = report;
 
-  return [...(resultStats && Object.entries(resultStats)), ...(diagnosticStats && Object.entries(diagnosticStats))]
```
nicolearagao: Help me understand what went on here:  the function is not safe if either object is null or undefined ?

nicolearagao (approved):
Nice! 👍🏼

src/views/scans/showAggregateReportModal.tsx
cdcabrera: We did...
- Reason... since we ended up using a single loop to do the component display we merged the two lists
- Considered... we considered keeping them separate and running them each through a support function which would support an iteration where dual columns are displayed... something like

```
const formatFilterSort = (arr): [string, unknown][] =>
    arr
      .filter(...
      .sort(...
      .map(...);

  const results = formatFilterSort(RESULT_OBJ);
  const diagnostics = formatFilterSort(DIAGNOSTIC_OBJ);

  return {
    hasResults: results.length > 0,
    hasDiagnostics: diagnostics.length > 0,
    results,
    diagnostics
  };
```

src/views/scans/__tests__/showAggregateReportModal.test.tsx
cdcabrera: yes, technically you could do handle this inside your API hook or the calling component...

> we reasoned not doing this because...
> -  we attempted to match existing API hooks where data is minimally transformed... generally in an attempt to broadcast a notification
> - separating it out of the component to help testing and creating a component that "just handles display". similar to dependency injection

The final reasoning went like...

In the end we saw 2 paths `before calling it`... the context of calling the function 
- in regards to the API
- in regards to a standalone unit that does something

Both paths end up relying on the function, ergo the function has the check

src/views/scans/showAggregateReportModal.tsx
cdcabrera: So this was an actual bug in implementation that we corrected in subsequent review fix... It leverages `truthiness` and the explicit setting of a value

- `undefined` falls back to the default object, ie `const { results: resultStats = {}`
- `null` is an explicit value that overrides the default empty object

You can see the end results yourself in console... 

```
var testing = [...(null && Object.entries(null))];
console.log(testing);
// result error, `not iterable`
```

vs

```
var testing = [...Object.entries(null || {})];
console.log(testing);
// result empty array/list
```

ruda (approved):

src/views/scans/showAggregateReportModal.tsx
```diff
@@ -0,0 +1,120 @@
...
+      onClose={() => onClose()}
+      {...(actions && { actions })}
+    >
+      <Title className="pf-v5-u-mb-lg" headingLevel="h2" size="md">
+        {t('modal.subtitle', { context: 'scan-id', value: report?.id })}
+      </Title>
+      <DescriptionList isHorizontal isFluid isCompact>
+        {stats.map(([key, value]) => {
+          return (
+            <DescriptionListGroup key={key}>
```
mirekdlugosz, line 101: Can you add `data-ouia-id` attribute here? Ideally value of attribute would be id from `aggregate` endpoint response, like `results.ansible_hosts_all` or `diagnostics.inspect_result_status_success`. I think this is what `key` represents right now.

This will help UI tests to grab labels and values. Right now we have to rely on group position or human-visible label - both of which might change on a whim.

mirekdlugosz:
Adding new item here makes [DISCOVERY-846](https://issues.redhat.com/browse/DISCOVERY-846) much worse than it was before. Now on the first click you can't even read the whole label, and second click moves table significantly - see screenshots.

I am a bit surprised this is not shown on your screenshots, as over here this is easily reproducible - any browser, full screen or not, temporary profile or not, using build container or `npm run start`.


| First click  | Subsequent clicks |
| ------------- | ------------- |
| Screenshot From 2025-05-28 13-48-48  | Screenshot From 2025-05-28 13-48-54  |

src/views/scans/showAggregateReportModal.tsx
mirekdlugosz: @cdcabrera thanks, that will help :)

PR #632, opened 2025-05-20
ci(konflux): adjust pipeline to support discovery release workflow

bruno-fs:
Relates to JIRA: DISCOVERY-969

Worflow for tagging and opening PR against other branches was tested on yaka repo.


infinitewarp (approved):
Basically same as https://github.com/quipucords/quipucords/pull/2953 which means this looks good to me! :shipit:

PR #637, opened 2025-05-29
fix(actionMenu): dsc-846 expose popperProps

cdcabrera-test:
## What's included
- fix(actionMenu): dsc-846 expose popperProps

## Notes
- quick patch that expose popperProps that help with alignment of the current actionmenu component
- apply the new exposed prop on all views... sources, scans, creds
- because of the way popper applies itself this may break at smaller screen sizes

## Updates issue/story
[DISCOVERY-846](https://issues.redhat.com/browse/DISCOVERY-846)


mirekdlugosz (approved):
That works well. Thank you.

List of actions is displayed below and to the left of kebab icon, as seen in screenshot below. Clicking kebab multiple times does not move the entire table.

Screenshot From 2025-05-29 17-47-07

Speaking of screen sizes, it seems that there are problems in browsers that have width below 768px. This is OK for now.

src/components/actionMenu/actionMenu.tsx
```diff
@@ -19,14 +26,16 @@ type Action<T> = {
 interface ActionMenuProps<T = unknown> {
   item: T;
   actions: Action<T>[];
+  popperProps?: Partial<PopperProps>;
```
nicolearagao, line 29: @cdcabrera David, if answering this won't take too much of your time, what would be the ideal fix for this use case ? You can just point me to the right component and I can take a look. Thank you.

PR #644, opened 2025-07-16
fix(en.json): strings files should not include function calls

abellotti:
- Strings file (i.e. public/locales/en.json) should not include function calls to define some values. Static strings should be just that, it's okay to have duplicates.

Issue: https://issues.redhat.com/browse/DISCOVERY-1027

## Summary by Sourcery

Bug Fixes:
- Remove function calls from public/locales/en.json and replace them with plain strings


public/locales/en.json
```diff
@@ -251,7 +251,7 @@
     "label_status_tooltip_success_sources_other": "{{count}} Successful Authentications",
     "label_status_tooltip_unreachable": "{{count}} Unreachable System",
     "label_status_tooltip_unreachable_other": "{{count}} Unreachable Systems",
-    "label_summary": "$t(modal.title_scan-summary)",
+    "label_summary": "Scan summary",
```
bruno-fs, line 254: while this works, there's the matter of the (unnecessary) duplication now.

but that's a whole different conversation, and since @nicolearagao mentioned we also need to cleanup this to remove unused strings, we can postpone this conversation.

infinitewarp (approved):

mirekdlugosz (approved):

PR #647, opened 2025-07-22
Improve unit test coverage

mirekdlugosz:
This is just enough work to get DISCOVERY-836 fixed.

## Before

coverage-before

## After

coverage-after

(Note that some rows changed color without changing numbers. That's because tool uses relative group - so file that had "good" coverage might fall to "not so good" level if **other** files improve.

I can use similar approach I used in `src/views/credentials/__tests__/addCredentialModal.test.tsx` to improve coverage in add source modal and scan modals. But I want to have confirmation this is right direction before doing further work.


package.json
```diff
@@ -45,7 +45,7 @@
     "test:lint-fix": "npm run test:lint -- --fix",
     "test:local": "jest --roots=./src --watch",
     "test:spell-support": "cspell ./README.md ./config/README.md ./CONTRIBUTING.md './scripts/*.js' './config/*.js' './tests/*.ts*' --config ./config/cspell.config.json",
-    "test:spell": "cspell './public/locales/**/en*json' './public/index.html' './src/**/*.ts*' --config ./config/cspell.config.json",
+    "test:spell": "cspell './public/locales/**/en*json' './public/index.html' './src/**/*.ts*' '!src/views/credentials/__tests__/__snapshots__/addCredentialModal.test.tsx.snap' --config ./config/cspell.config.json",
```
mirekdlugosz, line 48: I tried adding that to `cspell.config.json`, but apparently config ignores are applied **before** command-line selects, so it did not work. Gemini suggest to move path selection to config, and then you have everything in one place and ignore will work. I guess we can't fully do that because we want separate commands for two groups of files.

That snapshot contains partly-typed words, and that is what triggers cspell failures.

src/components/refreshTimeButton/__tests__/refreshTimeButton.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+  beforeEach(() => {
+    jest.useFakeTimers();
+    jest.setSystemTime(new Date(MOCKED_CURRENT_TIME));
+    const originalMomentUtcOffset = moment.prototype.utcOffset;
+    jest.spyOn(moment.prototype, 'utcOffset').mockImplementation(function (this: moment.Moment, offset: unknown) {
+      if (offset !== undefined) {
+        return originalMomentUtcOffset.apply(this, [offset]);
+      }
+      return 0;
```
mirekdlugosz, line 19: This was suggested by Gemini and I do admit I don't know what this conditional is needed. If I make that function `return 0` no matter the arguments, I get failing tests.

In our case we call `utcOffset` twice, and output of first call is input to second call, so I think we fundamentally use `0` every time anyway.

src/components/viewLayout/viewLayout.tsx
```diff
@@ -48,7 +48,12 @@ const AppLayout: React.FC<AppLayoutProps> = ({
   const Header = (
     <Masthead>
       <MastheadToggle>
-        <Button variant="plain" onClick={() => setSidebarOpen(!sidebarOpen)} aria-label="Global navigation">
+        <Button
+          variant="plain"
+          onClick={() => setSidebarOpen(!sidebarOpen)}
+          aria-label="Global navigation"
+          ouiaId="global-navigation"
```
mirekdlugosz, line 55: This is button without text content and I needed _something_ to find it in unit test.

src/components/viewLayout/__tests__/viewLayoutNavGroup.test.tsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+import { shallowComponent } from '../../../../config/jest.setupTests';
+jest.mock('../../../routes', () => {
```
mirekdlugosz, line 3: As far as I understand, this is canonical way of mocking a module - you have to create your mock **before** it is imported. Here, we want to test `ViewLayout`, which imports `routes` directly, so we need to mock routes first.

All that to test the path of menu with navigation hierarchy. We never use navigation hierarchy and I seriously doubt we ever will, so maybe we should just remove the support for it - we can always bring it back from git, if needed. But I already removed another module and decided that this time I will try to test it.

src/components/viewLayout/__tests__/viewLayoutToolbarInteractions.test.tsx
```diff
@@ -0,0 +1,81 @@
...
+      platform: { machine: 'x86_64' }
+    })
+  })
+}));
+import { AppToolbar as ViewToolbar } from '../viewLayoutToolbar';
+
+describe('ViewToolbar interactions', () => {
+  beforeEach(async () => {
+    // eslint-disable-next-line
+    const consoleSpy = jest.spyOn(console, 'log').mockImplementation(() => {});
```
mirekdlugosz, line 21: Opening modal or selecting logout [triggers `console.log()` call](https://github.com/quipucords/quipucords-ui/blob/main/src/components/viewLayout/viewLayoutToolbar.tsx#L69). I don't think this is intentional and something we should be testing - looks more like leftover from development time. But if I remove these `console.log()` calls, then TypeScript fails to build, complaining about unused variables...

src/components/viewLayout/viewLayoutToolbar.tsx
```diff
@@ -131,6 +131,7 @@ const AppToolbar: React.FC<AppToolbarProps> = ({ useLogout = useLogoutApi, useUs
                       variant="plain"
                       onClick={() => setHelpOpen(prev => !prev)}
                       isExpanded={helpOpen}
+                      data-ouia-component-id="help_menu_toggle"
```
mirekdlugosz, line 134: Same here - that's a button without text, but I need to click it in unit test.

src/views/credentials/__tests__/__snapshots__/addCredentialModal.test.tsx.snap
```diff
@@ -49,16 +49,394 @@ exports[`AddCredentialModal should render a basic component: basic 1`] = `
...
+    "ssh_key",
+    "s",
+  ],
+  [
+    "ssh_key",
+    "ss",
+  ],
+  [
+    "ssh_key",
+    "ssh",
```
mirekdlugosz, line 100: I'm seriously confused here.

We attach `handleInputChange` to `onChange` event, which triggers on every keystroke.  Our handler should pass HTMLElement **value**, so when I type "discovery" I would expect to see "d", "di", "dis" etc.

But this is not what I see for username - each event receives just a new character. I thought this is related to us using `useCallback` and react hooks (I think); I thought that react loop is nor properly set up in unit test, so these characters are never added to input. Not ideal, but acceptable.

But then for SSH key and password it **does** trigger event with string typed so far? So for these fields, it can add characters to input?

Now I think about it, SSH key is textarea, and password is input with "password" type, and maybe that makes a difference. But I would expect all of them to behave consistently.

src/views/credentials/__tests__/useCredentialsQuery.test.ts
```diff
@@ -0,0 +1,46 @@
...
+    isError: false,
+    error: null
+  }))
+}));
+import { useServiceQuery } from '../../../helpers/queryHelpers';
+import { type CredentialType } from '../../../types/types';
+import { useCredentialsQuery } from '../useCredentialsQuery';
+
+type CredentialsColumnKey = 'name' | 'type' | 'auth_type' | 'sources' | 'updated' | 'actions';
+type CredentialsSortableColumnKey = 'name' | 'type';
```
mirekdlugosz, line 16: These two are copy-pasted from [`useCredentialsQuery.ts`](https://github.com/quipucords/quipucords-ui/blob/main/src/views/credentials/useCredentialsQuery.ts#L14), because that module does not export them. I think it should. I can rework the module to export them, and then here I would import them instead of defining.

src/views/credentials/__tests__/addCredentialModal.test.tsx
```diff
@@ -194,15 +210,159 @@ describe('CredentialForm', () => {
...
+    const user = userEvent.setup();
+
+    await user.click(screen.getByText('Save'));
+
+    expect(container.innerHTML).toMatchSnapshot('onSave');
+  });
+
+  it('should submit updated credential data', async () => {
+    const mockOnSubmit = jest.fn();
+    const credential = {
```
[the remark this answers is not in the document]
nicolearagao: Good point. We could create a mockCredential factory with default values and override only the fields we need (like id and name) to keep tests concise. I also agree that separating v1 and v2 types would make the model clearer and avoid mixing unnecessary fields.

src/views/credentials/__tests__/useCredentialsQuery.test.ts
nicolearagao: Let's export them to avoid duplicates in tests

Action item: let's check with cory what should be the best approach here, concentrating all in one file (like type per example) or making it a bit more modular.

src/views/credentials/__tests__/addCredentialModal.test.tsx
```diff
@@ -129,31 +145,31 @@ describe('CredentialForm', () => {
 
   it('should render specific to authType for type network', async () => {
     const mockUseForm = jest.fn();
-    mockUseForm.mockReturnValue({ authType: 'SSH Key' });
+    mockUseForm.mockReturnValue({ authType: 'SSH Key', typeValue: 'network' });
```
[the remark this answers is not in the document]
nicolearagao: non-blocking: Maybe we need to check if typeValue is properly initialized or passed down from the modal. It could be worth refactoring so that CredentialForm doesn’t rely on external components to set typeValue, but instead handles a default internally when it’s missing.

src/views/credentials/__tests__/__snapshots__/addCredentialModal.test.tsx.snap
nicolearagao: I have no idea lol I messed with it a bit today but could not figure it out. I will look at it again tomorrow.

package.json
mirekdlugosz: 1. I created DISCOVERY-1045 to cover that. Generally I agree, but I would probably do that outside of this PR, and it seems like a good task to pick if someone wants to start contributing to this repo.
2. Fixed in 44eba6d4055776d18c9d4c544ddf3e41e8f38928

src/components/refreshTimeButton/__tests__/refreshTimeButton.test.tsx
mirekdlugosz: You are right, the problem is that `utcOffset` without arguments is just instance attribute getter, but with arguments is attribute setter; these instances are often modified through a chain of methods, so setters need to return modified instance.

I managed to simplify that even further in d615ecff2557e0d516abf8cdd77dcc8d36463258
(I can't make it arrow function because arrow functions can't take `this` argument)

src/components/viewLayout/viewLayout.tsx
mirekdlugosz: You are right. I wouldn't think of finding element by **aria** using getBy**Role**.

I would like to keep ouiaId in case we ever want to interact with that element from Camayoc, and to reinforce the idea that ouiaId are important in this project and we should overuse them than underuse them.

src/components/viewLayout/__tests__/viewLayoutToolbarInteractions.test.tsx
mirekdlugosz: Yes, that was the code that I was referring to, where I was getting errors if I just removed `console.log` calls.

I don't know why, but when I switched `value` to `_value`, suddenly eslint doesn't like that both arguments are underscored (it did not complain when only one of them was):

```
src/components/viewLayout/viewLayoutToolbar.tsx
  66:5  error  '_event' is defined but never used  @typescript-eslint/no-unused-vars
  67:5  error  '_value' is defined but never used  @typescript-eslint/no-unused-vars
  73:5  error  '_event' is defined but never used  @typescript-eslint/no-unused-vars
  74:5  error  '_value' is defined but never used  @typescript-eslint/no-unused-vars
```

I needed to also modify eslintrc file, to make this error go away.

Everything is included in e180fa083c11c38e6237ba02a2aa5f25ae9efd39

src/views/credentials/__tests__/useCredentialsQuery.test.ts
mirekdlugosz: Export is done in 5241404ead142d211f8d54f55ef89420b021a6ab 

I created DISCOVERY-1048 to track "use centralized or modular, but be consistent".

src/components/viewLayout/viewLayout.tsx
nicolearagao: fine with that!

PR #649, opened 2025-07-25
fix(useScanApi): increase page size on scan jobs API call

bruno-fs:
Updated the API call in useGetScanJobsApi to include a page_size parameter, ensuring that a maximum of 1000 scan jobs are retrieved per request.

Relates to JIRA: DISCOVERY-818

## Summary by Sourcery

Include a page_size parameter in useGetScanJobsApi to fetch up to 1000 scan jobs per request

Enhancements:
- Add page_size=1000 query parameter to the scan jobs API call in useGetScanJobsApi

Tests:
- Add unit test to verify axios.get is called with page_size=1000 parameter

### DIFF
--- a/src/hooks/useScanApi.ts
+++ b/src/hooks/useScanApi.ts
@@ -262,7 +262,7 @@ const useGetScanJobsApi = (onAddAlert: (alert: AlertProps) => void) => {
 
   const apiCall = useCallback(
     (scanId: number): Promise<AxiosResponse<ScanJobsResponse>> =>
-      axios.get(`${process.env.REACT_APP_SCANS_SERVICE}${scanId}/jobs/`),
+      axios.get(`${process.env.REACT_APP_SCANS_SERVICE}${scanId}/jobs/`, { params: { page_size: 1000 } }),
     []
   );
 
--- a/src/hooks/__tests__/__snapshots__/useScanApi.test.ts.snap
+++ b/src/hooks/__tests__/__snapshots__/useScanApi.test.ts.snap
@@ -425,6 +425,11 @@ exports[`useGetScanJobsApi should attempt an api call to retrieve a scanJob: api
 [
   [
     "/api/v1/scans/1/jobs/",
+    {
+      "params": {
+        "page_size": 1000,
+      },
+    },
   ],
 ]
 `;


src/hooks/useScanApi.ts
```diff
@@ -259,10 +259,9 @@ const useDeleteScanApi = (onAddAlert: (alert: AlertProps) => void) => {
 
 const useGetScanJobsApi = (onAddAlert: (alert: AlertProps) => void) => {
   const { t } = useTranslation();
-
```
mirekdlugosz, line 262: I think that if you keep that line, then you won't need any changes in `tests/__snapshots__/code.test.ts.snap`. That other file is common source of merge conflicts.

Does linter complain about this empty line?

src/hooks/__tests__/useScanApi.test.ts
```diff
@@ -328,6 +328,19 @@ describe('useGetScanJobsApi', () => {
...
+
+  it('should call axios with page_size=1000 query parameter', () => {
+    const { apiCall } = hookResult;
+    const spyAxios = jest.spyOn(axios, 'get');
+
+    const scanId = 123;
+    apiCall(scanId);
+
+    expect(spyAxios).toHaveBeenCalledTimes(1);
+    expect(spyAxios).toHaveBeenCalledWith(`${process.env.REACT_APP_SCANS_SERVICE}${scanId}/jobs/`, {
```
mirekdlugosz, line 340: Non-blocking: Personally, I think `process.env` does not belong to assertions. This quickly goes in the direction that variable equals itself, instead of checking the externally observable effects.

As far as I can tell, we don't really have established pattern for these situations. We hardly do any assertions around outgoing data. The only other reference to `process.env` in tests is a mock in [`useLoginApi.test.ts`](https://github.com/quipucords/quipucords-ui/blob/6f50e1fabf7a1cdf9215884193fe44e1119fa38a/src/hooks/__tests__/useLoginApi.test.ts#L217).

mirekdlugosz (approved):

src/hooks/__tests__/useScanApi.test.ts
bruno-fs: good call, I will use the proper value (I was too lazy to see what it was )

mirekdlugosz: Yeah, I think I had the same problem some time ago - when I run `eslint`, it does not complain, but when I run `eslint --fix`, it changes file content. I have strongly-worded opinions about this.

bruno-fs: Fixed. but turns out there is a snapshot test covering this. I thought this was uncovered because initially I did not saw that failure.
https://github.com/quipucords/quipucords-ui/commit/4af00502ef3925eb257fdf41610550bc56ec806a

nicolearagao (approved):

mirekdlugosz (approved):

PR #652, opened 2025-08-04
chore(deps): vendor react-table-batteries and the required piece of lib-ui

bruno-fs:
Vendor react-table-batteries in its totality and a small piece from
@migtools/lib-ui required for it.

The @migtools/lib-ui code is apache 2 licensed, with its LICENSE file
sitting next to
src/vendor/react-table-batteries/hooks/storage/useStorage.ts (the
only file vendored from that lib).

For some reason typescript can properly resolve all types used
on the vendored dependencies, while eslint parser had ~25 false
import errors. Maybe this is related to the fact that the eslint
and TS versions in use are not compatible (there's even a warning
about that when running `npm run test:lint`.

Furthermore all vendored code is being ignored from testing coverage.
If not ignored, the vendored could would have plummeted our current
coverage percentage.

Finally, vendoring these libraries resolved the dependency hell with
libraries like @tanstack/react-query (see [1] and [2]). This removes
the need of using the "--legacy-peer-deps" option on npm install during
the production build. This ocurred because all conflicts with incompatible libraries
were solved. Because of the resolution of this problem, the option
"overriding" axios package on "package.json" is not required anymore.

Relates to JIRA: DISCOVERY-798
Relates to JIRA: DISCOVERY-1035

[1]: https://github.com/quipucords/quipucords-ui/blob/6f50e1fabf7a1cdf9215884193fe44e1119fa38a/package-lock.json#L18
[2]: https://github.com/quipucords/quipucords-ui/blob/6f50e1fabf7a1cdf9215884193fe44e1119fa38a/package-lock.json#L3318

## Summary by Sourcery

Vendor the react-table-batteries library along with a required lib-ui storage hook, update imports to use vendored code, adjust tooling to ignore vendored files, and remove legacy peer dependency handling.

Enhancements:
- Add use-deep-compare dependency
- Remove custom axios override from package.json

Build:
- Remove --legacy-peer-deps flag from Containerfile to streamline npm installs

CI:
- Ignore src/vendor directory in ESLint and cspell configurations
- Exclude vendored code from Jest coverage reports

Chores:
- Vendor react-table-batteries and necessary @migtools/lib-ui useStorage hook to eliminate external dependency conflicts
- Update import statements to reference vendored react-table-batteries code


mirekdlugosz (approved):
Commit f72ecf49b8790bf4ad48fc5c89c43ceaf48bb73b looks alright. I trust commit 584840f48508626a2206daff740d56e466405e5b is accurate.

I see Camayoc passed for this PR, which gives me confidence there are no regressions.

I also tried running this locally (using `npm run start:using-server` from branch). I checked that main views look the same as in current `main`. I also tried all the actions (in action menu), buttons in rows, sorting, filtering, selecting many items and using buttons at the top.

I haven't noticed anything concerning.

nicolearagao (approved):

PR #658, opened 2025-08-07
TypeAhead: when input is used for filtering, dropdown is always visible

nicolearagao:
### Why
This started as a fix for a bug where the dropdown would not close after selecting an item from the search results.
While investigating, I found the component was maintaining two sources of truth for selected items:

- External selectedOptions (from props)

- Internal selected state

This duplication led to:

- State desynchronization and inconsistent dropdown behavior

- A “maximum update depth exceeded” error caused by effects trying to sync the two states in a loop

- Extra complexity that made the code harder to maintain

### What

- Removed internal selected state in favor of selectedOptions from props

- Updated all selection logic (onSelect, isSelected, placeholder text) to use selectedOptions

- Added justSelectedRef to prevent the menu from reopening right after a selection

- Ensured dropdown closes correctly in selection mode

Improved filtering logic to:

- Keep selected items at the top

- Sort remaining items alphabetically

- Display a "No results" option when applicable
[Screencast from 2025-08-17 22-39-51.webm](https://github.com/user-attachments/assets/0ea99568-3a1a-45b8-a35f-048d082851ab)

Relates to JIRA: DISCOVERY-1014

## Summary by Sourcery

Consolidate selection state to use only the controlled selectedOptions prop, fix dropdown visibility and filtering behavior, and cover new scenarios with tests

Bug Fixes:
- Remove duplicate internal selected state to avoid sync loops and inconsistent dropdown behavior
- Prevent menu from reopening immediately after selecting an item and ensure dropdown closes when appropriate

Enhancements:
- Simplify selection logic by relying solely on selectedOptions prop
- Keep selected items at the top of the filtered list and sort remaining options alphabetically
- Display a "No results found" entry when a filter yields no matches

Tests:
- Add tests for single-selection closing behavior, multi-selection flow, and "No results found" message

### DIFF
--- a/src/components/typeAheadCheckboxes/__tests__/typeaheadCheckboxes.test.tsx
+++ b/src/components/typeAheadCheckboxes/__tests__/typeaheadCheckboxes.test.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { render, screen } from '@testing-library/react';
+import { render, screen, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { shallowComponent } from '../../../../config/jest.setupTests';
 import { TypeaheadCheckboxes } from '../typeaheadCheckboxes';
@@ -15,7 +15,6 @@ describe('TypeaheadCheckboxes', () => {
 
   it('should clear selections and call onChange with [] when clear button is clicked', async () => {
     const mockOnChange = jest.fn();
-
     render(
       <TypeaheadCheckboxes
         options={[
@@ -31,5 +30,89 @@ describe('TypeaheadCheckboxes', () => {
     await userEvent.click(clearButton);
 
     expect(mockOnChange).toHaveBeenCalledWith([]);
+  });
+
+  it('should filter options, select one, and close dropdown when maxSelections is 1', async () => {
+    const onChange = jest.fn();
+
+    render(
+      <TypeaheadCheckboxes
+        options={[
+          { value: 'alpha', label: 'Alpha' },
+          { value: 'beta', label: 'Beta' }
+        ]}
+        selectedOptions={[]}
+        onChange={onChange}
+        maxSelections={1}
+      />
+    );
+
+    const input = screen.getByRole('combobox');
+    await userEvent.type(input, 'alp');
+
+    const alphaOption = await screen.findByText('Alpha');
+    await userEvent.click(alphaOption);
+
+    expect(onChange).toHaveBeenCalledWith(['alpha']);
+
+    // Wait for the menu to disappear
+    await waitFor(() => expect(screen.queryByRole('menu')).not.toBeInTheDocument());
+  });
+
+  it('should filter options, select multiple, and call onChange twice (controlled)', async () => {
+    const user = userEvent.setup();
+    const options = [
+      { value: 'alpha', label: 'Alpha' },
+      { value: 'beta', label: 'Beta' },
+      { value: 'gamma', label: 'Gamma' }
+    ];
+
+    /**
+     *
+     */
+    const Wrapper = () => {
+      const [selected, setSelected] = React.useState<string[]>([]);
+      return (
+        <TypeaheadCheckboxes options={options} selectedOptions={selected} onChange={setSelected} maxSelections={2} />
+      );
+    };
+
+    render(<Wrapper />);
+
+    const input = screen.getByRole('combobox');
+    await user.type(input, 'a');
+
+    const alphaOption = await screen.findByText('Alpha');
+    await user.click(alphaOption);
+
+    // Re-open the dropdown before selecting 'Beta'
+    await user.click(input);
+    await user.clear(input);
+    await user.type(input, 'bet');
+    const betaOption = await screen.findByText('Beta');
+    await user.click(betaOption);
+
+    // Now both should be selected
+    expect(screen.getByPlaceholderText('2 items selected')).toBeInTheDocument();
+
+    // Ensure the dropdown closes when maxSelections is reached
+    await waitFor(() => expect(screen.queryByRole('menu')).not.toBeInTheDocument());
+  });
+
+  it('shows "No results found" when filter yields no matches', async () => {
+    render(
+      <TypeaheadCheckboxes
+        options={[
+          { value: 'alpha', label: 'Alpha' },
+          { value: 'beta', label: 'Beta' }
+        ]}
+        selectedOptions={[]}
+      />
+    );
+
+    const input = screen.getByRole('combobox');
+    await userEvent.type(input, 'zzz');
+
+    expect(screen.getByText(/No results found/i)).toBeInTheDocument();
   });
 });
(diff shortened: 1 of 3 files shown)


src/components/typeAheadCheckboxes/__tests__/typeaheadCheckboxes.test.tsx
```diff
@@ -10,4 +12,88 @@ describe('TypeaheadCheckboxes', () => {
     const component = await shallowComponent(<TypeaheadCheckboxes {...props} />);
     expect(component).toMatchSnapshot('basic');
   });
+
```
nicolearagao, line 15: what I wanted to add here : a test that could verify the dropdown is closed when we click outside of it. The reason I couldn't: When testing the "click outside to close" behavior of the TypeaheadCheckboxes component, I encountered a limitation with our current testing setup using jsdom and React Testing Library. PatternFly's Select component, which our component relies on, uses portals and attaches event listeners to the real DOM to detect clicks outside the dropdown. However, jsdom does not fully emulate browser event propagation or portal behavior, so simulated clicks outside the component in our tests do not trigger the dropdown to close as they would in a real browser. As a result, our test for this specific behavior fails, even though the functionality works correctly when manually tested in the browser. To avoid confusion and false negatives in our test suite, I have chosen to skip this test.

src/components/typeAheadCheckboxes/__tests__/typeaheadCheckboxes.test.tsx
```diff
@@ -10,4 +12,88 @@ describe('TypeaheadCheckboxes', () => {
...
+    const options = [
+      { value: 'alpha', label: 'Alpha' },
+      { value: 'beta', label: 'Beta' },
+      { value: 'gamma', label: 'Gamma' }
+    ];
+
+    /**
+     *
+     */
+    const Wrapper = () => {
```
nicolearagao, line 54: This test verifies that the TypeaheadCheckboxes component works correctly in controlled mode when allowing multiple selections. In the test, I wrap the component in a Wrapper that manages the selected options using React state. We simulate a user typing to filter the options, selecting "Alpha", then re-opening the dropdown, filtering, and selecting "Beta". After both selections, we check that the placeholder updates to "2 items selected", confirming that both options are selected. Finally, we assert that the dropdown menu closes automatically when the maximum number of selections (maxSelections=2) is reached. This test ensures that the component properly supports controlled selection, updates the UI, and enforces the selection limit as expected.

src/components/typeAheadCheckboxes/typeaheadCheckboxes.tsx
```diff
@@ -196,12 +205,11 @@ const TypeaheadCheckboxes: React.FC<TypeaheadCheckboxesProps> = ({
           data-ouia-component-id="credentials_list_input"
         />
         <TextInputGroupUtilities>
-          {selected.length > 0 && (
+          {selectedOptions.length > 0 && (
             <Button
               variant="plain"
               onClick={() => {
                 setInputValue('');
-                setSelected([]);
```
mirekdlugosz, line 204: Blocking: "clear credentials" button is now broken, it does not do anything. It used to de-select everything that was selected.

It seems like something we could cover in unit test.

mirekdlugosz (changes requested):
I'm selecting "Request change" due to one blocking comment - clear button no longer works.

Overall, I'm on the fence. I appreciate the refactoring work and I like the direction this is moving. But looking at DISCOVERY-1014, that specific issue is still present - if I type something, I need to remove everything to hide a list. As long as anything is present in input field, the list is always going to be visible.

Now, since selecting item clears the input field content, the risk of encountering that issue is smaller. In current `main`, I type something, select my credentials and need to remove input content to proceed. Now it only really happens if I type something, **change my mind** and not select anything and would like to proceed.

So, I'm on the fence because we are not really fixing the issue - we are making it less likely for anyone to encounter it.

src/components/typeAheadCheckboxes/typeaheadCheckboxes.tsx
```diff
@@ -38,52 +38,52 @@ const TypeaheadCheckboxes: React.FC<TypeaheadCheckboxesProps> = ({
...
 
-      // Open the menu when the input value changes and the new value is not empty
-      if (!isOpen) {
+      if (!isOpen && !justSelectedRef.current) {
         setIsOpen(true);
       }
     }
 
-    setSelectOptions(newSelectOptions);
+    const selectedOptionsSorted = filteredOptions.filter(option => selectedOptions.includes(option.value));
```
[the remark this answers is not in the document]
nicolearagao: You were absolutely right - there was no .sort() call on this variable, so the name was misleading. The new name accurately reflects that it contains filtered (not sorted) selected options.

src/components/typeAheadCheckboxes/typeaheadCheckboxes.tsx
```diff
@@ -38,52 +38,52 @@ const TypeaheadCheckboxes: React.FC<TypeaheadCheckboxesProps> = ({
...
+    const unselectedOptions = filteredOptions
+      .filter(option => !selectedOptions.includes(option.value))
+      .sort((a, b) => a.label.localeCompare(b.label));
+
+    const newOptions = [...selectedOptionsSorted, ...unselectedOptions];
+
+    setSelectOptions(prev => {
+      const prevValues = prev.map(opt => opt.value).join(',');
+      const nextValues = newOptions.map(opt => opt.value).join(',');
+      return prevValues !== nextValues ? newOptions : prev;
```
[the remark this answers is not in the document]
nicolearagao: The original optimization wasn't primarily about performance, it was preventing an infinite loop. The loop was not a regression, but it was making it impossible for me to write any tests, except for the basic test we have, it would always explode on my face. 
The new approach I am proposing does it a bit more efficiently by avoiding the overhead of string operations while achieving the same goal.

src/components/typeAheadCheckboxes/typeaheadCheckboxes.tsx
nicolearagao: ah, sorry. THis is fixed on the other PR by adding onChange([]) to properly clear all selections (rather than trying to use setSelected). The test I added covers this behavior, I am rebasing with main, the behavior should be restored.

src/components/typeAheadCheckboxes/__tests__/typeaheadCheckboxes.test.tsx
nicolearagao: Created the follow up story to track this work: https://issues.redhat.com/browse/DISCOVERY-1080

src/components/typeAheadCheckboxes/__tests__/typeaheadCheckboxes.test.tsx
nicolearagao: ahhhh, I see now ! I downloaded the video you shared, I will open a slack thread to address it.

PR #659, opened 2025-08-07
second bug

nicolearagao:


mirekdlugosz (approved):
Works as expected :+1: 
And thanks for adding a test!

A question: I see with that change, we call `onChange` in all the same places where we call `setSelected`, **except** [`useEffect` body](https://github.com/quipucords/quipucords-ui/blob/43b80122a57614cffea6ff837cf87ae7ab56bcad/src/components/typeAheadCheckboxes/typeaheadCheckboxes.tsx#L53). Should we add `onChange` there as well?

mirekdlugosz:
> @mirekdlugosz answering your question: No, we shouldn't call onChange inside this effect even if we don't remove it. The purpose of this effect is to update the internal state when the parent changes the selectedOptions prop. Calling onChange here would create a feedback loop: the parent updates selectedOptions then the effect runs and sets state then onChange is called then parent updates selectedOptions again, and so on.

I see. Thanks!

One day I really need to sit down and understand these React hooks and effects.

PR #661, opened 2025-08-14
fix(sourceConnection): reset state on close

mirekdlugosz:
Relates to JIRA: DISCOVERY-820

## Summary by Sourcery

Reset the expanded state of the ShowConnectionsModal when it is closed and verify this behavior with a new test.

Bug Fixes:
- Clear the expanded sections state when the modal is closed either via the onClose handler or the Close button

Tests:
- Add a test to ensure the expanded state is reset and the onClose callback is invoked on modal close

### DIFF
--- a/src/views/sources/__tests__/showSourceConnectionsModal.test.tsx
+++ b/src/views/sources/__tests__/showSourceConnectionsModal.test.tsx
@@ -1,3 +1,4 @@
+import '@testing-library/jest-dom';
 import React from 'react';
 import { screen, render } from '@testing-library/react';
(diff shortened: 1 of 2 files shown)


cmchase:
LGTM. Some purists might complain that resetting it "manually" to its default state is a bad idea and would push you towards https://react.dev/reference/react/useReducer, but this is a very usecase and it's well covered by tests.

cmchase:
I don't remember any hard rules on using both of them in the same file but it would make sense to stick to one pattern. If it's not elsewhere in the app then there's _definitely_ no reason to introduce it.

The changes look good!

PR #674, opened 2025-08-29
fix(views): Align number of cells in header and table

mirekdlugosz:
react-table-batteries creates cell with select box by default. By adding "selection" column to header, we ended up with one more cell than in body.

## Before

Screenshot From 2025-08-29 12-51-56

## After

Screenshot From 2025-08-29 12-52-05

## Summary by Sourcery

Remove selection column header cells in various list views to align the number of header and body cells

Bug Fixes:
- Fix misalignment of header and table body cells by removing the extra selection column in credentials, scans, and sources list views

Tests:
- Update snapshot to reflect removed selection column header in code.test.ts


nicolearagao (approved):
oops, that was a leftover from a different checkbox solution approach. Thanks for fixing.

PR #675, opened 2025-08-29
fix(rescan): disable action for running scans

mirekdlugosz:
The logic behind `canRequestRescan` is inspired by backend [`ScanJob.status_restart` implementation](https://github.com/quipucords/quipucords/blob/e18ecf5d39d7df3f63191353f154ac26c3242164/quipucords/api/scanjob/model.py#L364). In practice, there's no UI to pause a scan and scan would be pending only for extremely short period of time - most likely shorter than user can navigate from Sources to Scans view.

Also add conditional tooltips explaining why other actions are disabled. See screenshots below.

Also also change localization ids of tooltips displayed when delete action is disabled. These ids were misleading as they contained "edit" instead of "delete". I can drop this commit if smuggling refactors like that bothers someone.

## Running scan - summary report tooltip

Running scan - summary report tooltip

## Running scan - Rescan tooltip

Running scan - Rescan tooltip

## Running scan - Download tooltip

Running scan - Download tooltip

## Failed scan - summary report tooltip

Failed scan - summary report tooltip

## Failed scan - Download tooltip

Failed scan - Download tooltip

## Summary by Sourcery

Disable rescan, summary report, and download actions for scans that are not eligible and provide context-specific tooltips; update delete-tooltip localization keys for credentials and sources; add tests for scan action helpers.

Enhancements:
- Add canRequestRescan helper to determine when rescan is allowed
- Conditionally disable summary report, rescan, and download buttons based on scan status
- Display explanatory tooltips for disabled scan actions
- Rename tooltip localization keys for delete actions in credentials and sources lists

Tests:
- Add unit tests for canRequestRescan helper covering various scan statuses
- Add unit tests for canAccessMostRecentReport helper with all scan statuses


nicolearagao (approved):
nice!

src/helpers/helpers.ts
```diff
@@ -236,9 +236,19 @@ const getTitleImg = (isBrand = UI_BRAND) => ((isBrand && titleImgBrand) || title
 const canAccessMostRecentReport = (job?: scanJob | MostRecentScan) =>
   job?.status === 'completed' && job?.report_id !== undefined;
 
+/**
+ * Tell if a scan can be scanned again.
+ *
+ * @param {scanJob | MostRecentScan} job
+ * @returns {boolean}
+ */
+const canRequestRescan = (job?: scanJob | MostRecentScan) =>
```
cmchase, line 245: Observation: this is a clever use of `includes`. My gut says that the || OR operator isn't needed but after digging into it checking against a string as a fallback could be considered a best practice.

cmchase, line 245: Observation: I had to dig into `types` to see why we're doing a union on what _should_ be the same type at first glance. It looks like we have a note to refactor/merge those -- it's not something we should do now but it did catch my eye.

cmchase (approved):
I skimmed the code but my approval is for the language used.

src/helpers/helpers.ts
cmchase: Ahhh, very cool! Thanks for the insight -- I'm sure I would have gotten that code hint if I pulled this down to my IDE.

infinitewarp (approved):

PR #691, opened 2025-09-09
Properly isolate unit tests

mirekdlugosz:
One commit per area. Commit messages are copy-pasted so they can be understood in isolation.

- prefer `mockRestore()` over `mockClear()`. `mockClear()` only clears internal mock data, while keeping mock in place. Since we spy on global objects, we ended up putting spies on top of other spies.
- make sure that every spy calls `mockResolvedValue`, `mockRejectedValue` or `mockImplementation`, which ensures that spied function is never called.

See also internal docs PR 97.

Relates to JIRA: DISCOVERY-1101

## Summary by Sourcery

Restore and isolate unit tests by using jest.restoreAllMocks and providing explicit mock implementations for spies

Tests:
- Replace jest.clearAllMocks() with jest.restoreAllMocks() across tests to fully restore spied methods between tests
- Add explicit mockResolvedValue, mockRejectedValue, mockImplementation, or mockReturnValue for jest.spyOn() calls on axios, cookies, document APIs, and i18next to prevent real implementations from running

### DIFF
--- a/src/components/i18n/__test__/i18n.test.tsx
+++ b/src/components/i18n/__test__/i18n.test.tsx
@@ -13,7 +13,7 @@ describe('I18n', () => {
   });
 
   it('should attempt to update locale', async () => {
-    const spyI18next = jest.spyOn(i18next, 'changeLanguage');
+    const spyI18next = jest.spyOn(i18next, 'changeLanguage').mockResolvedValue(jest.fn() as any);
     const props = {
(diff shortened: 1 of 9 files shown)


nicolearagao (approved):
This looks good Mirek. They will conflict big time with the PRs I have opened, let's merge yours first, so I can also apply those patterns to the tests I am adding.

PR #693, opened 2025-09-10
feat(sources): honor source_task_status in sources list

abellotti:
Do not merge, this is a POC. Needs additional testing for different scenarios. single source, different source types, etc.

- Used for showing the source specific status for the last connection. Was incorrectly showing failed when the scan job that targeted multiple sources and another source failed, when the specific source shown was successful.

Requires: https://github.com/quipucords/quipucords/pull/3024

## Summary by Sourcery

Add support for source-specific connection task status in the sources list view

Enhancements:
- Add API constants and type fields for source_task_id and source_task_status
- Update sources list view to use source_task_status for status labels and icons


abellotti:
closing for now until we start working on DISCOVERY-1105

PR #702, opened 2025-09-16
remove normalizeTotal modulus

nicolearagao:
# Fix: Make normalizeTotal modulus optional

## Problem
Created 105 credentials to test source view pagination, but couldn't navigate to page 2 in credentials view to see all my test data. The `normalizeTotal` function was applying `count % 100` in development mode (105 → 5 total items), breaking pagination across all views.

## Solution
Removed the normalizeTotal helper function entirely and replaced all usages with direct data?.count || 0 access. This eliminates unnecessary complexity and fixes pagination issues in development mode by ensuring actual count values are always used.

## Changes
Remove normalizeTotal function and its tests from helpers.ts
Replace helpers.normalizeTotal(data) with data?.count || 0 in:
src/views/credentials/viewCredentialsList.tsx
src/views/scans/viewScansList.tsx
src/views/sources/viewSourcesList.tsx
Remove normalizeTotal from helpers module exports

### DIFF
--- a/src/helpers/__tests__/helpers.test.ts
+++ b/src/helpers/__tests__/helpers.test.ts
@@ -214,15 +214,6 @@ describe('normalizeCommaSeparated', () => {
   });
 });
 
-describe('normalizeTotal', () => {
-  it('should normalize total values', () => {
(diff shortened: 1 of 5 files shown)


mirekdlugosz (approved):
Good call :+1:

PR #705, opened 2025-09-18
Add Overview page

mirekdlugosz:
The page is available under `/overview` URL. The navigation menu item is hidden behind feature flag. Feature flag is off by default and can be opted in by setting `REACT_APP_FEATURE_OVERVIEW` environment variable to "true".

Even when present, "Overview" is NOT a default page after logging in.

The page is unfinished. The text is still being finalized. Last time I checked, it was unclear if lower left should have a text description or an image.

But since page is hidden by default, we can review the structure implementation and possibly merge it, and fill in the content later.

Screenshot From 2025-09-18 13-31-30


mirekdlugosz:
Looks like placeholder Latin in locale is failing integration test. So we can't merge. But we can review.

src/views/overview/viewOverview.css
```diff
@@ -0,0 +1,35 @@
...
+      var(--pf-t--global--border--color--default) calc(50% - 1px),
+      var(--pf-t--global--border--color--default) calc(50% + 1px),
+      transparent 50%
+    );
+    /* On CSS side, we can't know how many rows we have.
+     * The idea is that second number below is 1/n , where n is number
+     * of rows. So 50% for 2 rows, 33.33% for 3 rows etc. Then the line
+     * between items should be drawn for each row.
+     */
+    /*background-size: 100% 50%;*/
```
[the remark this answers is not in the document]
mirekdlugosz: You mean, like the comment right above it?

public/locales/en.json
```diff
@@ -373,7 +373,67 @@
     "label_credentials": "Credentials",
     "label_scans-names": "Scans: {{name}}",
     "label_sources": "Sources",
-    "label_not-found_home": "Take me home"
+    "label_not-found_home": "Take me home",
+    "overview": {
+      "title": "Overview",
+      "getting-started": {
```
mirekdlugosz, line 379: I didn't know you can have arbitrarily nested objects in locale file.

Personally, I find it much superior to semi-flat structure we used to have. For me, the choice between dash (`-`) and underscore (`_`) in existing keys always felt completely random. It seems the only established standards are to use dash instead of space, and i18next suggest to use underscore to separate plurals. I don't think we even follow that consistently.

The current solution is inconsistent with rest of the file and I will flatten it if people feel strongly about it. But personally I would use nested objects for all new strings, with long term goal to gradually port old flat structure to nested one.

cmchase: Agreed on the value and use. I like the chaining and hierarchy it provides.

src/views/overview/viewOverview.css
cmchase, line 20: Question: is this commented rule needed or can we delete it?

src/routes.tsx
```diff
@@ -22,6 +23,13 @@ interface IAppRoute {
 }
 
 const routes: IAppRoute[] = [
+  {
```
cmchase, line 26: Suggestion: I believe we're going to need to update line 67 (`path="/"`) to catch the default route to set this as the "home page" on first load.

src/views/overview/viewOverview.css
mirekdlugosz: I left it for now because it might be needed if we want to allow these elements to wrap. I especially want to hear feedback from people with Apple hardware. Unfortunately, I can't easily simulate that screen on my machine.

public/locales/en.json
```diff
@@ -373,7 +373,67 @@
...
+        "header": "Getting started resource",
+        "item1": {
+          "label": "About Discovery",
+          "content": "<0>Red Hat Discovery is a tool that inspects your environment to identify physical and virtual hosts, operating systems, and configuration data. It helps you understand your Red Hat subscription usage across your entire infrastructure.</0>"
```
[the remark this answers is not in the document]
mirekdlugosz: It's a placeholder for React component. Think HTML tag.

Most translations work well when you want to translate relatively short texts, like a word, line, paragraph. But what if you want to translate few paragraphs?

You can keep a document structure in TSX file and use translation keys like `overview.getting-started.item1.content.para-1`, `overview.getting-started.item1.content.para-2` etc. But that's awkward.

On the other hand, you could put HTML tags in translation file, but then your translation files drive the document structure, and that's also awkward. And it's going to be a maintenance nightmare when you decide to switch from `<p>` to `<div>` and have multiple language files where change is needed.

Current approach is a bit between. You can still use single key for multiple paragraphs, and you keep abstracted structure in translation files. The actual components are provided by TSX file, where they belong ([`components` attribute here](https://github.com/quipucords/quipucords-ui/blob/66c2e9e3cceeb43c2c6a614c87fd5c6f60c34d1c/src/views/overview/viewOverview.tsx#L57)).

It's a bit over the top for this revision, as all the strings fit in single paragraph. But I figured getting started cards and Q&A answers _might_ require more than single paragraph, or some rich text, so I decided to use it. On the other hand, I haven't used it for process steps, as these seem to value brevity.

infinitewarp (changes requested):
First of all, thank you Mirek for going the extra mile to add `{isDarkTheme ? overviewSecurityDarkSrc : overviewSecuritySrc}` to make the image swap dynamically. The image looks much better, more professional, and more cohesive now that it matches the theme. ❤️ 

Second, I'm only just now realizing, upon looking at the current branch served in my web browser, that all of this new text uses the downstream name "Discovery" in many places despite being served from upstream "Quipucords" here. Could we template the uses of "Discovery" throughout `public/locales/en.json` and fill it dynamically at runtime? I see elsewhere we have `helpers.UI_NAME` which is populated at startup using environment variables and is used by `aboutModal`, `errorMessage`, and `viewLayout`, some of which then set a variable like `uiName` which is later used to drop in the preverred name when building some of the components. Maybe we can borrow that idea here.

ruda (approved):

public/locales/en.json
```diff
@@ -373,7 +373,67 @@
...
+        "item2": {
+          "label": "Gather and add credentials",
+          "content": "Prepare the credentials for each of your sources. Some of them may require more than one."
+        },
+        "item3": {
+          "label": "Add your sources",
+          "content": "You are ready to connect your source now with the credentials from the last steps."
+        },
+        "item4": {
+          "label": "Run scans, download reports",
```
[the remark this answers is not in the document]
mirekdlugosz: Done in 2ffd3224e1e3a230a5d99d1d0434f7352f2d9203

nicolearagao (approved):

PR #715, opened 2025-10-07
fix(overview): don't leak icon color change to other views

mirekdlugosz:
This fixes a following bug:
1. Have some credentials and sources. Ideally with successful scans.
2. Open Overview page (default after logging in). Refresh page.
3. Open Sources page using left menu

Result: Icon in "Last Connected" column has changed color. It is blue instead of green.
Refreshing a page on Sources view returns color back to normal.

The root cause is that CSS is split and loaded conditionally as components are mounted, but it is not unloaded later - so all component CSS is global, but not necessarily loaded right after refresh.

Before:
Screenshot From 2025-10-07 12-44-56


nicolearagao (approved):

mirekdlugosz:
@cmchase I'm not sure if I understand.

`pf-v6-c-icon__content` class is for element that contains only `<svg>`. It doesn't affect text. HTML structure for our buttons in affected place looks like that:

```html
<button class="pf-v6-c-button pf-m-link pf-m-small">
    <span class="pf-v6-c-button__text">
        <span class="pf-v6-c-icon">
            <span class="pf-v6-c-icon__content pf-m-success">
                <svg .../>
            </span>
        </span> 
        Last connected 5 hours ago
    </span>
</button>
```

PatternFly docs are not great on how the library is supposed to be used, or at least I haven't seen docs covering that. My understanding is that current iteration is extensively using CSS variables. There's a lot of selectors that only assign a new value to some variable - and some other place uses that variable to affect how things are displayed.

That's also what caused this bug. There's actually a CSS selector with **higher** specificity that says "assign green to this variable". But since I changed CSS to say "use some other variable as value for `color`", that other CSS rule changed value of variable that was not used anymore.

So now I follow more closely to what PatternFly expects, at least I think so - I assign a new value to variable. Then on other screen, another CSS rule overrides the same variable with new value.

Hope that makes sense. I'm going to merge the PR now to not block a release - we can continue on Slack or other venues.

PR #923, opened 2026-04-16
feat: Ask for Lightspeed authentication status after logging in

mirekdlugosz:
After logging in, send request for Lightspeed authentication status. Response that token is 'valid' is understood as "user is logged in". Anything else ('missing', 'expired', 'invalid', 4xx or 5xx response) is understood as "user is not logged in", without notification.

Relates to JIRA: DISCOVERY-1299

## Summary by Sourcery

Add a reusable Lightspeed authentication status check and invoke it after successful login when the reports feature is enabled.

New Features:
- Trigger a Lightspeed authentication status request immediately after a successful login to determine if the user is authenticated with Lightspeed.

Enhancements:
- Extract a shared helper for querying the Lightspeed authentication status and reuse it both in polling logic and post-login checks.


nicolearagao (approved):
LGTM! Nice refactor pulling currentLightspeedAuthStatus out so it's reused between polling and the new login check. One minor note: the catch block in requestLightspeedAuthStatus doesn't log the error via console.error like other catch blocks in the hooks do, but given this is a best effort restore it's fine as is.

mirekdlugosz:
@nicolearagao FYI, I added `console.error` call behind `helpers.TEST_MODE` conditional - just like we have in other places.

Since you mentioned this is minor, I force-pushed (instead of fixup) and I'm going to merge this soon.

PR #929, opened 2026-04-27
feat(credentials): Vault secret path auth for OpenShift and Ansible

nicolearagao:
Adds a third authentication type for OpenShift and Ansible (AAP) credentials: 
- Vault secret path, backed by vault_secret_path and optional vault_mount_point. The option is always shown but disabled when no global HashiCorp Vault configuration exists.
-  Submit logic sends only the fields that match the selected auth type (mutually exclusive with token / username+password).

User-visible behavior
- Authentication type dropdown includes Vault for OpenShift and Ansible only (unchanged for Network and types without this dropdown).
- Selecting Vault shows Secret path (required) and Mount point (optional; placeholder indicates default secret)..
- Save stays disabled until required fields are valid; standard inline errors for empty required path.
- Vault availability is determined by GET on the Vault config endpoint (via useGetHashicorpVaultConfigApi when the add/edit credential modal is open for OpenShift or Ansible).

## Manual testing (Vault credentials — OpenShift + Ansible)

**Setup:** `GET /api/v2/auth/hashicorp-vault/` returns **404** (no config) or **200** (config present).

**Without Vault config (404):** For OpenShift and Ansible, **Vault secret path** is disabled in the auth menu; **Token** / **username+password** still create and save normally.

**With Vault config (200):** **Vault secret path** is enabled; choosing it shows only **Secret path** (required) and **Mount point** (optional); other auth inputs hidden; **Save** blocked until secret path is filled; create sends vault fields only.

**Edit:** Vault cred loads path/mount; switching auth types and saving works; list shows correct auth type.

**Regression:** Other credential types unchanged; sources can still attach OpenShift/Ansible credentials including Vault-backed ones.

## Summary by Sourcery

Add Vault secret path as a new authentication option for OpenShift and Ansible credentials and wire it to a global HashiCorp Vault configuration check.

New Features:
- Introduce Vault secret path authentication type for OpenShift and Ansible credentials, with optional Vault mount point and conditional availability based on global Vault config.

Enhancements:
- Extend credential form logic and dropdown component to support disabled menu items with tooltips and to cleanly submit only fields relevant to the selected auth type.
- Add helper mappings and API constants to recognize Vault-based credentials and expose global HashiCorp Vault configuration via a dedicated React Query hook.

Tests:
- Add unit and snapshot tests for the Vault configuration hook and for credential forms covering Vault-enabled and Vault-disabled scenarios, as well as new getCleanedFormData behavior.


abellotti (approved):
UI code looks good.

I also ran the UI and trying the vault secret path and mount points with both OpenShift and Ansible (wasn't presented in other types). Worked great.  Also tested against as server that did not have the HashiCorp Vault configured.  Got the correct error propagated back. So 👍

src/components/simpleDropdown/simpleDropdown.tsx
```diff
@@ -75,14 +77,28 @@ const SimpleDropdown: React.FC<SimpleDropdownProps> = ({
...
-              {item}
-            </DropdownItem>
-          ))}
+          dropdownItems.map(({ item, ouiaId, isDisabled, disabledTooltip }) => {
+            const useNativeDisabled = Boolean(isDisabled) && !disabledTooltip;
+            return (
+              <DropdownItem
+                key={item}
+                isDisabled={useNativeDisabled}
+                isAriaDisabled={isDisabled}
```
mirekdlugosz: Do we need this `useNativeDisabled`? We never pass values through `Boolean` in our codebase (well, there are few places where we use `.filter(Boolean)`). `isDisabled` is already a boolean value, isn't that enough?

Is there a reason to have possibly two different values passed to `isDisabled` vs `isAriaDisabled`? Shouldn't both get the same (whether it's `useNativeDisabled` or `isDisabled`).

Last thing, I **think** `isAriaDisabled` will be enough here. We use it exclusively in [actionMenu](https://github.com/quipucords/quipucords-ui/blob/344bc115b5713d0e19b7a23ecb14e2ac7203d540/src/components/actionMenu/actionMenu.tsx#L70). As far as I remember, isAriaDisabled displays tooltip on hover and blocks clicking (so you don't need `if (!isDisabled)` in `onClick`).

src/locales/en.json
```diff
@@ -274,7 +274,8 @@
     "credentials": {
       "add-modal": {
         "auth_type": {
-          "label": "Authentication Type"
+          "label": "Authentication Type",
+          "vault_disabled_tooltip": "To use Vault authentication, a HashiCorp Vault server must first be configured in the application settings."
```
mirekdlugosz:
```suggestion
          "vault_disabled_tooltip": "To use Vault authentication, a HashiCorp Vault server must be configured first."
```

"in the application settings" suggests this is something you can do in UI, but you can't - at least not now.

src/views/credentials/addCredentialModal.tsx
```diff
@@ -76,6 +93,21 @@ const getCleanedFormData = (
...
     });
 
+  if (authType === helpers.authType.VaultSecretPath) {
+    if (!String(cleanedData.vault_mount_point ?? '').trim()) {
+      delete cleanedData.vault_mount_point;
+    }
+    (['username', 'password', 'become_user', 'become_method', 'become_password'] as const).forEach(key => {
+      const v = cleanedData[key];
+      if (v === '' || v === undefined) {
+        delete cleanedData[key];
```
mirekdlugosz: These fields are very similar to default `maskedFields`, and the logic seems to duplicate `Object.entries(cleanedData)` few lines above.

I don't understand why this is being done, but can't we modify one of `maskedFields` / `fieldsAllowedToBeEmpty` and let rest of the function do the work as-is? Instead of having another list of field names?

src/views/credentials/addCredentialModal.tsx
```diff
@@ -87,8 +119,10 @@ const deriveAuthType = (credential?: Partial<CredentialResponse>, typeValue?: st
     case 'openshift':
     case 'rhacs':
       return 'Token';
+    case 'ansible':
+      return helpers.authType.UsernameAndPassword;
     default:
-      return 'Username and Password';
+      return helpers.authType.UsernameAndPassword;
```
mirekdlugosz: If `helpers.authType.UsernameAndPassword` is already `default`, do we need a special case for `ansible`?

src/views/credentials/addCredentialModal.tsx
```diff
@@ -141,7 +177,8 @@ const useCredentialForm = ({
         username: !!credential?.username,
         name: !!credential?.name,
         become_user: !!credential?.become_user,
-        become_method: !!credential?.become_method
+        become_method: !!credential?.become_method,
+        vault_secret_path: !!(credential?.vault_secret_path && String(credential.vault_secret_path).trim())
```
mirekdlugosz: Shouldn't we have `vault_mount_point` here as well? Seems we have a map of all fields.

src/views/credentials/addCredentialModal.tsx
```diff
@@ -397,18 +471,58 @@ const CredentialForm: React.FC<CredentialFormProps> = ({
...
             dropdownItems={
               (typeValue === 'network' && [
-                { item: 'Username and Password', ouiaId: 'password' },
-                { item: 'SSH Key', ouiaId: 'ssh_key' }
-              ]) || [
-                { item: 'Token', ouiaId: 'auth_token' },
-                { item: 'Username and Password', ouiaId: 'password' }
-              ]
+                { item: helpers.authType.UsernameAndPassword, ouiaId: 'password' },
+                { item: helpers.authType.SSHKey, ouiaId: 'ssh_key' }
```
mirekdlugosz: Maybe these two should be moved to `networkAuthDropdownItems`?

It's a bit weird that we extracted OpenShift and Ansible to standalone variables, and left network inlined.

src/views/credentials/addCredentialModal.tsx
```diff
@@ -614,6 +728,7 @@ const AddCredentialModal: React.FC<AddCredentialModalProps> = ({
       onClose={() => onClose()}
     >
       <CredentialForm
+        isOpen={isOpen}
```
mirekdlugosz: I don't love how `CredentialForm` gets a new prop, which needs to be synchronized with parent `Modal`. I think something like `onOpen` event would be better suited here.

I'm not 100% sure, but it seems you can use `useEffect(() => { }, [])` (that is, effect without any dependencies) to a have a code running only on component mount. And I **think** `CredentialForm` will be mounted when modal appears, and unmounted when it disappears.

src/hooks/useGetHashicorpVaultConfigApi.ts
```diff
@@ -0,0 +1,47 @@
...
+    }
+    if (!helpers.TEST_MODE) {
+      console.error(error);
+    }
+    return { vaultConfigured: false };
+  }
+};
+
+const useGetHashicorpVaultConfigApi = (enabled: boolean) =>
+  useQuery({
```
mirekdlugosz, line 38: It seems there's only one other place where we use `useQuery` (in query helper, it helps us to set up table data). Is there a reason to use it instead of one-off axios request, which we use in bunch of other places?

src/hooks/__tests__/useGetHashicorpVaultConfigApi.test.ts
```diff
@@ -0,0 +1,55 @@
...
+  });
+
+  it('should not call the API when disabled', () => {
+    const spyAxios = jest.spyOn(axios, 'get');
+    renderHook(() => useGetHashicorpVaultConfigApi(false), { wrapper: createWrapper() });
+    expect(spyAxios.mock.calls).toMatchSnapshot('apiCall, disabled');
+  });
+
+  it('should retrieve vault global config on success', async () => {
+    const spyAxios = jest.spyOn(axios, 'get').mockResolvedValue({
```
mirekdlugosz: Please prefer `mockResolvedValueOnce` in new code. Without `Once`, data is returned for all subsequent request. With `Once`, you need to return data for every request - this helps to ensure that a number of requests made matches your expectation. Especially important for requests that are made in response to reactive changes, which are sometimes surprising.

src/views/credentials/__tests__/addCredentialModal.test.tsx
```diff
@@ -3,9 +3,30 @@ import { render, renderHook, screen } from '@testing-library/react';
...
+  isPending: false,
+  isFetching: false,
+  isSuccess: true
+}));
+
+jest.mock('../../../hooks/useGetHashicorpVaultConfigApi', () => ({
+  useGetHashicorpVaultConfigApi: (enabled: boolean) => mockVaultConfigQuery(enabled)
+}));
+
+beforeEach(() => {
```
mirekdlugosz, line 21: Is it OK to have a mock in `beforeEach` without mockRestore in `afterEach`?

src/views/credentials/__tests__/addCredentialModal.test.tsx
```diff
@@ -496,6 +522,89 @@ describe('CredentialForm', () => {
...
+    await user.click(toggle as Element);
+  };
+
+  it('should render OpenShift auth menu with Vault secret path disabled', async () => {
+    const user = userEvent.setup();
+    render(<CredentialForm isOpen credentialType="openshift" onSubmit={jest.fn()} />);
+
+    await openAuthTypeMenu(user);
+
+    // Dropdown list is portaled outside RTL `container`; snapshot the open menu only.
```
mirekdlugosz: `RTL`?

I know it might mean "right-to-left" when referring to content flow in horizontal direction, but here it seems to mean something else. I did not find `RTL` pattern in our codebase.

src/views/credentials/__tests__/addCredentialModal.test.tsx
```diff
@@ -496,6 +522,89 @@ describe('CredentialForm', () => {
...
+
+    await user.type(screen.getByPlaceholderText(/view.credentials.add-modal.name.placeholder/), 'ocp-cred');
+    await user.type(screen.getByPlaceholderText(/view.credentials.add-modal.token.placeholder/), 'my-token');
+    await user.click(screen.getByText(/view.credentials.add-modal.actions.save/));
+
+    expect(mockOnSubmit).toHaveBeenCalledTimes(1);
+    const payload = mockOnSubmit.mock.calls[0][0];
+    const redacted = {
+      ...payload,
+      ...(payload.auth_token ? { auth_token: '<redacted>' } : {}),
```
mirekdlugosz: I don't subscribe to "never use logic in tests" world view, but this is unit test that, as far as I can tell, uses static data. Shouldn't we already know if `payload.auth_token` is truthy or not?

mirekdlugosz (commented):
I was not able to test this against working Vault instance, and I did not regression-test non-Vault scenarios. I trust they all work.

There are few things in the code that I found puzzling. I'm not saying they are wrong - probably I am just missing some things. I left comments, please push back.

src/components/simpleDropdown/simpleDropdown.tsx
nicolearagao: Keeping the current useNativeDisabled split: with native disabled alone, the Vault row’s tooltip didn’t show on hover. Rows with disabledTooltip use isAriaDisabled + tooltipProps without native disabled so hover still works. I’ve added a brief comment in simpleDropdown explaining that. Happy to revisit simplification after a quick browser pass if we want to match actionMenu exactly.

src/views/credentials/__tests__/addCredentialModal.test.tsx
nicolearagao: The Vault hook is a jest.fn() from jest.mock, not a spy. jest.restoreAllMocks() doesn’t reset that the way it does jest.spyOn.
I reset it by a file-level beforeEach that sets the default mockImplementation before every test. The “vault disabled” tests use an extra beforeEach in that describe to override, no leak into other tests because the global beforeEach runs first on the next test. So afterEach isn’t needed for this mock here.

src/views/credentials/addCredentialModal.tsx
nicolearagao: The loop over Object.entries only deletes empty fields that are listed in maskedFields. For Vault we need to omit other keys (username, become_user, etc.) when they’re empty so they aren’t sent as "", which the generic loop never removes. The list is Vault-specific on purpose so we don’t add those keys to maskedFields globally and change behavior for other auth types.
I decided to refactor by doing three things: naming the Vault-only field list so its scope is obvious, extracting a small helper that deletes keys whose values are empty or undefined, and reusing that helper for both the masked-field path (once we know which masked keys are empty) and the Vault path, so behavior stays the same but the flow is a bit clearer. Cursor suggested a bunch of really ellaborate refactors, I decided to not go there as I am not confident it's the right thing to do.

src/views/credentials/addCredentialModal.tsx
nicolearagao: hm, I am not sure about this one. 
CredentialForm only uses it to set enabled on useGetHashicorpVaultConfigApi (fetch Vault config when the modal is open and the type is OpenShift/Ansible), so the API isn’t called while the modal is closed.
On useEffect with an empty dependency list: that only matches “runs every time the user opens the modal” if CredentialForm actually unmounts when the modal closes. With many PatternFly modals, children stay mounted when isOpen is false, so a mount effect runs once, not on each open and I’d still need some visibility signal, which is what isOpen provides today.
On onOpen: I’d still be wiring the same “modal is open” fact into the hook; it’s mostly a different API surface, not necessarily less coupling, unless I restructure (for example only rendering the form when open).
Conditional {isOpen && <CredentialForm … />} could let me drop the prop, but that’s a slightly bigger behavioral change (remounting / state) and deserves explicit UX testing so I’d treat that as a follow-up.
Given that, passing isOpen from the parent is a small, explicit tradeoff and keeps Vault query gating correct and easy to reason about.
I am voting to keep it as it is for now.

src/hooks/useGetHashicorpVaultConfigApi.ts
nicolearagao: I used React Query so the Vault config fetch can be enabled only when the modal needs it, and so the result is cached briefly (same queryKey, staleTime) if the user opens the modal again without hammering the endpoint. A one-off axios + useState would work too; this avoided extra effect/state boilerplate. Happy to switch style if we prefer matching other one-off API calls.

src/hooks/__tests__/useGetHashicorpVaultConfigApi.test.ts
nicolearagao: right, we did talk about it some time ago.

src/views/credentials/__tests__/addCredentialModal.test.tsx
nicolearagao: oh, React Testing Library. I updated the comment.

src/views/credentials/__tests__/addCredentialModal.test.tsx
nicolearagao: Good point. These tests use fixed user flows, so the payload shape is known up front, there’s no need for conditional redaction. A slip through the cracks in my Cursor workflow.

src/views/credentials/__tests__/addCredentialModal.test.tsx
mirekdlugosz: Thanks for explanation.

src/hooks/useGetHashicorpVaultConfigApi.ts
mirekdlugosz: I don't think I have very strong preference either way. I was mostly curious. `useQuery` / React Query looks pretty useful - maybe we should use it in more places, instead of axios?

Thanks for your answer. I think we can keep it this way.

src/views/credentials/addCredentialModal.tsx
mirekdlugosz: I wonder if we could just remove these files in `case helpers.authType.VaultSecretPath` branch, instead of having additional ` if (authType === helpers.authType.VaultSecretPath)` few lines later?

Also, does Vault need to care about stuff like `ssh_key` and `ssh_passphrase`? You can't use Vault for network credential.

I don't want to push too hard either way. I generally really dislike this part of our code - it's trying to handle too much of business logic. But we should not refactor it completely here, that would be a scope creep.

So, please think a little about my two questions above. But if you think we should just continue with what we have right now, I would be fine with that.

src/components/simpleDropdown/simpleDropdown.tsx
mirekdlugosz: >  I’ve added a brief comment in simpleDropdown explaining that.

I don't see any comment added to this file. Did you forget to push some commit?

Anyway, I tried this locally:

```typescript
              <DropdownItem
                key={item}
                isAriaDisabled={isDisabled}
                onClick={() => {
                  onSelect(item);
                }}
                ouiaId={ouiaId}
                tooltipProps={isDisabled && disabledTooltip ? { content: disabledTooltip } : undefined}
              >
                {item}
              </DropdownItem>

```

And with Vault not configured, it produces:

Image

Tooltip is displayed on hover. I can't click the item (so there is no action triggered).

Isn't this exactly what we want here? I don't understand what is the benefit of the more complex code in this PR.

src/views/credentials/addCredentialModal.tsx
mirekdlugosz: > On useEffect with an empty dependency list: that only matches “runs every time the user opens the modal” if CredentialForm actually unmounts when the modal closes. With many PatternFly modals, children stay mounted when isOpen is false, so a mount effect runs once, not on each open and I’d still need some visibility signal, which is what isOpen provides today.

Did you try it?

I added this inside `const CredentialForm` (around line 397):

```
  useEffect(() => {
    console.log(`${JSON.stringify(new Date())} jestem efektem w formularzu`);
  }, []);
```

And I see new entry is added in console log every time I open "Add Credential" modal. Entry is not added when I close it.

Vault would be a bit more complex due to requirement of state mamangement, but otherwise seems doable.

I don't love how useEffect is rather non-intuitive here, so I would probably also add a comment "run this when component mounts (modal is opened)". I consider that better than current duplicated `isOpen`, but not great enough to make it a blocker. All I ask is that you consider this approach.

nicolearagao: I tested this and you were right: CredentialForm remounts on each modal open, so passing isOpen was unnecessary. I removed that prop and now gate the Vault config query only by type (openshift/ansible). I reran the credential modal tests and they pass.

src/views/credentials/addCredentialModal.tsx
nicolearagao: I made a small cleanup without changing behavior:
I moved the Vault-only cleanup into the Vault case in the switch, so that logic is in one place.
I still clear ssh_key and ssh_passphrase there, because tests show those values can still be present from earlier auth selections and should not be sent with Vault.
So this is simpler flow, same behavior, and tests still pass.

src/components/simpleDropdown/simpleDropdown.tsx
nicolearagao: huh, this is weird. When I was going through your last review I did try to simplify it and the tooltip did not appear while the item was disabled, which is why I introduced the native/ARIA split initially. After your follow-up and another pass, I realized the key is to avoid native isDisabled on the DropdownItem and rely on isAriaDisabled with tooltipProps; with that setup, I can reproduce the expected behavior (tooltip on hover, no action on click).

mirekdlugosz: No problem. Glad we arrived to agreement here :) .

src/views/credentials/addCredentialModal.tsx
mirekdlugosz: I see we don't even need `useEffect`. A bit unexpected, but if it works, then it works. I tested manually that request is sent when I open a modal. Then I need to wait for 2 minutes to get another request.

Thanks.

mirekdlugosz (approved):
Thanks for replying to all the comments. I still have not tested this against actual vault instance, but the code looks much better now!

PR #937, opened 2026-05-06
feat(creds): gate vault auth behind feature flag

nicolearagao:
Hide HashiCorp Vault credential authentication behind a feature flag (REACT_APP_FEATURE_VAULT_AUTH), defaulting to false. When disabled, the "Vault secret path" auth type is removed from OpenShift and Ansible credential dropdowns and the vault config API call is skipped entirely.

## Summary by Sourcery

Gate HashiCorp Vault credential authentication behind a new feature flag and hide related options when the feature is disabled.

New Features:
- Introduce a FEATURE_VAULT_AUTH feature flag to control availability of HashiCorp Vault credential authentication.

Enhancements:
- Conditionally fetch HashiCorp Vault configuration and show Vault secret path auth type in credential forms only when the feature flag is enabled.

Build:
- Add REACT_APP_FEATURE_VAULT_AUTH build argument to the Containerfile to configure the Vault auth feature flag at build time.

Tests:
- Update credential form snapshots to reflect conditional display of the Vault secret path auth option based on the feature flag.


mirekdlugosz (approved):
Looks good. Also tested local server with and without a feature flag. With flag turned off, there is no Vault API call and item is not visible in dropdown. With flag on, item is visible in dropdown (disabled in my case), and there is vault API request in dev tools network tab.

PR #950, opened 2026-05-21
fix(scans): hide deep scan options for non-network sources

infinitewarp:
This change fixes a years-old misleading UI behavior that wrongly implied that the "deep scan" options applied to all source types.

Deep scan options (JBoss EAP/Fuse/WS checkboxes and alternate directories field) are only relevant to the backend API for network-type sources. Previously, these options appeared for all source types, causing confusion when creating scans for satellite, vcenter, openshift, rhacs, or ansible sources.

The scan modal now checks if at least one network source is included and only displays the deep scan section when applicable. This works for both single-source and multi-source scan creation flows.

Assisted-by: Claude Sonnet 4.5 <noreply@anthropic.com>

## Demo video

https://github.com/user-attachments/assets/f78abcdc-ed72-4e59-bb23-8b889f72ddc8

## Summary by Sourcery

Restrict deep scan options in the source scan modal to network-type sources and adjust related UI behavior accordingly.

Bug Fixes:
- Prevent deep scan checkboxes from appearing for non-network sources in the scan creation modal.
- Hide the search directories field when deep scan options are not applicable to the selected sources.

Tests:
- Add tests covering deep scan visibility for network, non-network, mixed, and undefined source selections, including conditional display of the search directories field.

### DIFF
--- a/src/views/sources/__tests__/addSourcesScanModal.test.tsx
+++ b/src/views/sources/__tests__/addSourcesScanModal.test.tsx
@@ -42,22 +42,6 @@ describe('AddSourceModal', () => {
 });
 
 describe('Conditional Deep Scan Display', () => {
-  const mockConnection = {
-    end_time: '2023-01-01T00:00:00Z',
-    id: 1,
-    report_id: 1,
-    source_systems_count: 1,
-    source_systems_failed: 0,
-    source_systems_scanned: 1,
-    source_systems_unreachable: 0,
-    start_time: '2023-01-01T00:00:00Z',
-    status: 'completed',
-    status_details: { job_status_message: 'Success' },
-    systems_count: 1,
-    systems_scanned: 1,
-    systems_failed: 0
-  };
-
   it('should show deep scan options when scanning a network source', async () => {
     const networkSources = [
       {
@@ -68,7 +52,6 @@ describe('Conditional Deep Scan Display', () => {
         hosts: ['host1'],
         exclude_hosts: [],
         credentials: [],
-        connection: mockConnection,
         ssl_cert_verify: false,
         disable_ssl: false
       }
@@ -94,7 +77,6 @@ describe('Conditional Deep Scan Display', () => {
         hosts: ['satellite.example.com'],
         exclude_hosts: [],
         credentials: [],
-        connection: mockConnection,
         ssl_cert_verify: true,
         disable_ssl: false
       }
@@ -118,7 +100,6 @@ describe('Conditional Deep Scan Display', () => {
         hosts: ['host1'],
         exclude_hosts: [],
         credentials: [],
-        connection: mockConnection,
         ssl_cert_verify: false,
         disable_ssl: false
       },
@@ -130,7 +111,6 @@ describe('Conditional Deep Scan Display', () => {
         hosts: ['satellite.example.com'],
         exclude_hosts: [],
         credentials: [],
-        connection: { ...mockConnection, id: 2 },
         ssl_cert_verify: true,
         disable_ssl: false
       }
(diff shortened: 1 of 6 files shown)


nicolearagao (approved):

src/views/sources/__tests__/addSourcesScanModal.test.tsx
```diff
@@ -41,6 +41,186 @@ describe('AddSourceModal', () => {
...
+    const networkSources = [
+      {
+        id: 1,
+        name: 'Network Source',
+        port: 22,
+        source_type: 'network',
+        hosts: ['host1'],
+        exclude_hosts: [],
+        credentials: [],
+        connection: mockConnection,
```
mirekdlugosz: `connection` is optional field (in backend response; UI types might disagree). We can remove this line and `mockConnection` object.

src/views/sources/__tests__/addSourcesScanModal.test.tsx
```diff
@@ -41,6 +41,186 @@ describe('AddSourceModal', () => {
...
+    ];
+
+    await act(async () => {
+      await render(<AddSourcesScanModal isOpen={true} sources={networkSources} />);
+    });
+
+    expect(screen.getByText(/deep-scan\.label/)).toBeInTheDocument();
+    expect(screen.getByText(/deep-scan\.jboss_eap/)).toBeInTheDocument();
+    expect(screen.getByText(/deep-scan\.jboss_fuse/)).toBeInTheDocument();
+    expect(screen.getByText(/deep-scan\.jboss_ws/)).toBeInTheDocument();
```
mirekdlugosz: I'm on the fence. This isn't wrong, and I like explicitness of these assertions. On the other hand, we rely on snapshots heavily and I feel we should use `expect().toMatchSnapshot()`. 

You can point agent at `src/components/secretInput/__tests__/secretInput.test.tsx` as an example of code where we use `render`, perform actions and still use snapshot tests. But we need to review snapshots to ensure they contain something we want.

mirekdlugosz (approved):
Conditionally ACK.

The code looks OK and I verified it works as advertised.   
But there are two things we could improve in tests. I don't think they are blocking, but at least removal of `mockConnection` should take no time at all.

src/views/sources/__tests__/addSourcesScanModal.test.tsx
infinitewarp: I pushed another commit (`test: replace explicit assertions with snapshots for deep scan display`) that switches these tests to use `expect(...).toMatchSnapshot(...)` instead.

src/views/sources/__tests__/addSourcesScanModal.test.tsx
infinitewarp: This is a good catch, but it required a little refactoring in another commit (`fix(types): make SourceResponse.connection optional to match API reality`) to make it work. Ultimately, this required making the `connection` attribute optional on the base `SourceResponse` class where it was _required_ (incorrectly) before and updating a couple existing tests to remove mock data that was never even being used (was apparently only there to satisfy the not-optional attribute). Even though that may be a bit more than I was expecting in this change, I think it's also more technically correct than what we've had.

mirekdlugosz: I did not know and I did not check, but I had a hunch that UI API types definitions might incorrectly claim `connection` is required. Thanks for taking care of this.

src/views/sources/__tests__/addSourcesScanModal.test.tsx
mirekdlugosz: I skimmed over snapshots and saw that some of them have deep scan part (and references to various JBosses), while others do not. I think this is sufficient for me.

Thanks.

PR #979, opened 2026-07-10
feat(reports): add scan ID and scan name columns, sorting, and filter

infinitewarp:
NOTE: This PR should not merge until AFTER https://github.com/quipucords/quipucords/pull/3240 merges because this branch depends on API changes being introduced in that PR's branch.

Improves the reports table to surface scan context alongside each report. This should make the page much more usable because without these the user has no information about where the listed reports came from.

- Adds `Scan ID` and `Scan Name` columns to the reports table
- Makes both new columns sortable alongside the existing ID, Origin, and Created columns
- Adds a `Scan ID` filter to the reports toolbar

This also includes some unrelated fixed for TypeScript config and pins `prettier` to `3.8.5` which seems like was missed in some previous commits.

Relates to JIRA: DISCOVERY-1343

## Summary by Sourcery

Enhance the reports list UI with additional scan context and improve supporting configuration.

New Features:
- Add Scan ID and Scan Name columns to the reports table and include them in the reports list view.
- Enable sorting by Scan ID and Scan Name alongside existing sortable report columns.
- Add a Scan ID search filter to the reports toolbar for narrowing reports by originating scan.

Enhancements:
- Update reports query column mappings and types to support optional scan_id and scan_name fields.
- Extend reports table type definitions to treat scan identifiers and names as nullable values.

Build:
- Refine Jest transform configuration to use ts-jest with an inline tsconfig that declares Jest types.
- Update package lock to pin the prettier dependency version.

Tests:
- Adjust useReportsQuery tests to cover the new scan-related columns and mappings.

## Summary by CodeRabbit

* **New Features**
  * Added **Scan ID** and **Scan name** columns to the reports table.
  * Expanded the reports toolbar with filters and placeholders for **Report Scan ID** and **Report Scan name**.
  * Updated table sorting to support ordering by **Scan ID** and **Scan name**.
  * Added localized “not available” text and placeholders (shown as **—** when missing).
* **Bug Fixes**
  * Improved handling of reports where scan details are absent, so missing values display correctly.


infinitewarp:
@sourcery-ai summary

infinitewarp:
@sourcery-ai summary

src/views/reports/viewReportsList.tsx
```diff
@@ -283,6 +293,8 @@ const ReportsListView: React.FunctionComponent = () => {
             {currentPageItems?.map((report: ReportType, rowIndex) => (
               <TrWithBatteries key={report.id} item={report} rowIndex={rowIndex}>
                 <Td columnKey="id">{report.id}</Td>
+                <Td columnKey="scan_id">{report.scan_id ?? '—'}</Td>
+                <Td columnKey="scan_name">{report.scan_name || '—'}</Td>
```
cmchase: Question: are we intentionally using nullish coalescence on 296 and logical OR on 297?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_OR

infinitewarp: Good catch! This inconsistency was due to two different agents implementing the same solution for the different fields in different ways. Looks like `??`/Nullish is more correct here because the API returns an actual null if these values are not set. Fixed in a new commit.

jest.config.js
```diff
@@ -31,7 +31,7 @@ module.exports = {
   testMatch: ['<rootDir>/**/__tests__/**/*.{ts,tsx}', '<rootDir>/**/*.{spec,test}.{ts,tsx}'],
   testEnvironment: 'jsdom',
   transform: {
-    '^.+\\.(js|jsx|ts|tsx)$': '<rootDir>/node_modules/ts-jest',
+    '^.+\\.(js|jsx|ts|tsx)$': ['<rootDir>/node_modules/ts-jest', { tsconfig: { types: ['jest'] } }],
```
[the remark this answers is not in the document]
infinitewarp: I don't want to update dependencies in this PR unless it's strictly required. We'll round up those updates elsewhere.

abellotti (approved):

